PR #35, opened 2014-10-30
Added joint limits to ChainIkSolverPos_LMA

facontidavide:
I have added joints limit to the IK.
I don't know if this follows strictly your coding standards (by the way, sorry for the different tabulation).

We might have created a new class ChainIkSolverPos_LMA_JL, but it would have been a lot of code replicated.

### DIFF
--- a/orocos_kdl/src/chainiksolverpos_lma.cpp
+++ b/orocos_kdl/src/chainiksolverpos_lma.cpp
@@ -73,7 +73,8 @@ ChainIkSolverPos_LMA::ChainIkSolverPos_LMA(
 	svd(6, _chain.getNrOfJoints(),Eigen::ComputeThinU | Eigen::ComputeThinV),
 	diffq(_chain.getNrOfJoints()),
 	q_new(_chain.getNrOfJoints()),
-	original_Aii(_chain.getNrOfJoints())
+    original_Aii(_chain.getNrOfJoints()),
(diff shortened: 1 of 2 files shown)


smits:
@facontidavide could you provide a version which is consistent with the existing API?

rethink-forrest:
@smits The API for all existing functions and constructors is consistent. Is there a reason not to add the four new functions which are fairly straight forward?

@facontidavide, my only suggestion would be to document the four new functions. Should only take 2-3 lines each.

rethink-forrest:
@smits Or are you asking that q_min and q_max be added to the constructor (maybe a second constructor)?  Guess it would make sense that use_joint_limits would be set to true if the limits were provided by the constructor.

jf---:
@rethink-forrest , yes, so my comment was about that --- for instance look at the constructor of ChainIkSolverPos_NR_JL. Why would you want a different constructor / API for the exact same functionality.

smits:
As a second thought I like the API of this solver more than the ChainIkSolverPos_NR_JL. I would propose to keep the functions for setting the limits here, add them in ChainIkSolverPos_NR_JL for consistency and add "initial" values in the constructor for ChainIkSolverPos_LMA for consistency with ChainIkSolverPos_NR_JL. This way we keep API backwards compatibility.

PR #37, opened 2014-11-13
Added support to RPATH for CMake >= 2.8.12

francesco-romano:
CMake 2.8.12 changed how `RPATH` is handled (at least for OS X). Furthermore in CMake 3.0 a warning is issued if these changes are not integrated (or the policy disabled).

Using `RPATH` avoids messing with the linker search path in environmental variables.

My patch should be backcompatible: `RPATH` is disabled by default and is supported only on CMake >= 2.8.12.
In case CMake is less recent or (by default) `RPATH` is not chosen, it behaves as before (so it keeps also the CMake warning).

Compiling and installing with `RPATH` on OS X gives the following results:
```bash
$ otool -D liborocos-kdl.dylib 
liborocos-kdl.dylib:
@rpath/liborocos-kdl.1.3.dylib
```
In this way the library is relocable (the library name does not contain the absolute path to the library).

Just one note in this PR:
Lines from 88 to 99 are needed only if KDL links to other non-system libraries (which currently it does not). But they should not harm if kept.

### DIFF
--- a/orocos_kdl/src/CMakeLists.txt
+++ b/orocos_kdl/src/CMakeLists.txt
@@ -74,7 +74,7 @@ IF(NOT (CMAKE_VERSION VERSION_LESS 2.8.12))
 
     IF(OROCOSKDL_ENABLE_RPATH)
         #Configure RPATH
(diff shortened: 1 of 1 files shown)


smits:
Looks ok to me, I'll merge after some additional tests.

PR #39, opened 2014-12-08
Check that the maxvel is obtainable in a half trapezoidal velocity profile

rethink-forrest:
Calculate the maximum velocity obtained during the half trapezoidal profile to calculate the correct timing. Prior to this change, the assert in path_composite.cpp that checks the path distance, s, failed if the max velocity was not reached.


smits:
I would be really careful with changing the behaviour of this kind of code. As these are the type of changes that easily break applications.

We probably have use cases for each behaviour (max velocity as the maximum velocity or as the end velocity), I would vote for a solution where we can support both.

rethink-forrest:
Updated header as requested for clarification.

PR #47, opened 2015-03-04
Use SolverI interface for chainidsolvers

snrkiwi:
The chain FK and IK solvers already use this interface, and this ensures that the chain ID solvers do as well.

Also contains a small fix to the SolverI interface to provide descriptions of all SolverI error codes.

### DIFF
--- a/orocos_kdl/src/solveri.hpp
+++ b/orocos_kdl/src/solveri.hpp
@@ -114,6 +114,8 @@ public:
 	{
 		if (E_NOERROR == error) return "No error";
 		else if (E_NO_CONVERGE == error) return "Failed to converge";
+		else if (E_UNDEFINED == error) return "Undefined value";
+		else if (E_DEGRADED == error) return "Converged but degraded solution";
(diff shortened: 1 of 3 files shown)


orocos_kdl/src/solveri.hpp
```diff
@@ -114,6 +114,8 @@ class SolverI
 	{
 		if (E_NOERROR == error) return "No error";
 		else if (E_NO_CONVERGE == error) return "Failed to converge";
+        else if (E_UNDEFINED == error) return "Undefined value";
+        else if (E_DEGRADED == error) return "Converged but degraded solution";
```
smits: could you fix these inconsistent whitespaces?

smits:
Except for the above whitespace inconsistencies, looks good to me.

snrkiwi:
On Mar 04, 2015, at 07:05, Ruben Smits <notifications@github.com> wrote:

> Except for the above whitespace inconsistencies, looks good to me.
> 
Done. Fix force pushed.
S

Issue #46, opened 2015-02-12
chainiksolvervel_pinv_nso math implementation incorrect


rethink-forrest:
I am working on a 7 DOF arm. With testing, chainiksolvervel_pinv_nso gives the exact same answer of chainiksolvervel_pinv every time.

Looking at the code, the nullspace math is not implemented properly on lines 106 - 123. There is no documentation for the attempted approach, and the reference paper from 1977 uses a different optimization without SVD decomposition.

As far as I can guess, the attempted technique is

t = weights * (q_opt - q_in)

q'_out += alpha * V_c * V_n^T * t

where V_c is the first 6 columns of V, and V_n are the remaining columns. Is this correct, and is there a paper or other references for this method?

If I know the desired method, I should be able to submit a patch quickly.
 
Other issues:
Line 110: The index i should start at jac.rows() and not jac.rows()+1. For a 7 dof arm, jac is 6x7, and this loop will never be run as written.  Same error is repeated in line 119.

Line 120: attempts to read tmp2 at index jac.rows()+1, but tmp2 only has jac.columns()-jac.rows() elements.

smits:
On Thu, Feb 12, 2015 at 7:09 PM, Forrest Rogers-Marcovitz <
notifications@github.com> wrote:

> I am working on a 7 DOF arm. With testing, chainiksolvervel_pinv_nso gives
> the exact same answer of chainiksolvervel_pinv every time.
>
> Looking at the code, the nullspace math is not implemented properly on
> lines 106 - 123. There is no documentation for the attempted approach, and
> the reference paper from 1977 uses a different optimization without SVD
> decomposition.
>

The paper does not explain how it calculates the generalized inverse of J,
this code uses an SVD to do that


> As far as I can guess, the attempted technique is
>
> t = weights * (q_opt - q_in)
>
> q'_out += alpha * V_c * V_n^T * t
>

According to the paper it should be q'_out += (I_n - G_2*J)*alpha*t

with G_2 = J^t(J*J^t)^-1 (aka pseudo inverse)

and (I_n-G_2*J) the projection of alpha*t into the null-space.

I have not looked into the math yet if the current code results in the same.


> where V_c is the first 6 columns of V, and V_n are the remaining columns.
> Is this correct, and is there a paper or other references for this method?
>
> If I know the desired method, I should be able to submit a patch quickly.
>
Other issues:
>
Line 110: The index i should start at jac.rows() and not jac.rows()+1. For
[...]

PR #50, opened 2015-04-08
Fixes to chainiksolvervel_pinv_nso

rethink-forrest:
Regarding #46 

Issues fixed:
* Errors in the how the matrices were iterated
    * Line 110: The index i should start at jac.rows() and not jac.rows()+1. For a 7 dof arm, jac is 6x7, and this loop will never be run as written. Same error is repeated in line 119.
    * Line 120: attempts to read tmp2 at index jac.rows()+1, but tmp2 only has jac.columns()-jac.rows() elements.

These two issues prevented any nullspace optimization from occurring - returning the same solution as  chainiksolvervel_pinv. I switched to Eigen matrices to improve reliability and readability in the math.

* The optimization function and the nullspace projection math did not appear to be correct.

Here is the new nullspace optimization math:

        // Given the cost function g, and the current joints q, desired joints qd, and weights w:
        // t = g(q) = 1/2 * Sum( w_i * (q_i - qd_i)^2 )
        //
        // The jacobian Jc is:
        //  t_dot = Jc(q) * q_dot
        //  Jc = dt/dq = w_j * (q_i - qd_i) [1 x nj vector]
        //
        // The pseudo inverse (^-1) is
        // Jc^-1 = w_j * (q_i - qd_i) / A [nj x 1 vector]
        // A = Sum( w_i^2 * (q_i - qd_i)^2 )
        //
        // We can set the change as the step needed to remove the error times a value alpha:
        // t_dot = -2 * alpha * t
        //
        // When we put it together and project into the nullspace, the final result is
        // q'_out += (I_n - J^-1 * J) * Jc^-1 * (-2 * alpha * g(q))

### DIFF
--- a/orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
+++ b/orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
@@ -20,19 +20,21 @@
 // Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "chainiksolvervel_pinv_nso.hpp"
+#include "utilities/svd_eigen_HH.hpp"
 
 namespace KDL
 {
     ChainIkSolverVel_pinv_nso::ChainIkSolverVel_pinv_nso(const Chain& _chain, JntArray _opt_pos, JntArray _weights, double _eps, int _maxiter, double _alpha):
         chain(_chain),
         jnt2jac(chain),
-        jac(chain.getNrOfJoints()),
-        svd(jac),
-        U(6,JntArray(chain.getNrOfJoints())),
-        S(chain.getNrOfJoints()),
-        V(chain.getNrOfJoints(),JntArray(chain.getNrOfJoints())),
-        tmp(chain.getNrOfJoints()),
-        tmp2(chain.getNrOfJoints()-6),
+        nj(chain.getNrOfJoints()),
+        jac(nj),
+        U(MatrixXd::Zero(6,nj)),
+        S(VectorXd::Zero(nj)),
+        Sinv(VectorXd::Zero(nj)),
+        V(MatrixXd::Zero(nj,nj)),
+        tmp(VectorXd::Zero(nj)),
+        tmp2(VectorXd::Zero(nj)),
         eps(_eps),
         maxiter(_maxiter),
         alpha(_alpha),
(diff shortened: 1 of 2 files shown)


orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
```diff
@@ -71,76 +74,86 @@ namespace KDL
...
-        for(i = 0; i < jac.columns(); i++)
-            tmp(i) = weights(i)*(opt_pos(i) - q_in(i));
-        
-        //Vtn*tmp
-        for (i=jac.rows()+1;i<jac.columns();i++) {
-            tmp2(i-(jac.rows()+1)) = 0.0;
-            for (j=0;j<jac.columns();j++) {
-                tmp2(i-(jac.rows()+1)) +=V[j](i)*tmp(j);
-            }
+        // Now onto NULL space
```
rethink-imcmahon: +1 I'm a big fan of explaining your math in the comments here

orocos_kdl/src/velocityprofile_traphalf.cpp
```diff
@@ -58,9 +58,9 @@ void VelocityProfile_TrapHalf::PlanProfile1(double v,double a) {
 	a3 = 0;
 	a2 = 0;
 	a1 = startpos;
-	b3 = a/2;
+	b3 = a/2.0;
```
rethink-imcmahon: Good catch on the `int` division...

orocos_kdl/src/path_composite.cpp
```diff
@@ -53,8 +53,8 @@ namespace KDL {
 // you propably want to use the cached_index variable
 double Path_Composite::Lookup(double s) const
 {
-	assert(s>=0);
-	assert(s<=pathlength);
+	assert(s>=-1e-12);
```
rethink-imcmahon: Can `1e-12` be made a constant somewhere?

rethink-imcmahon:
:shipit: 
Just based on my once-over, and reading ticket #46 this pull request seems like a solid addition. @rethink-forrest, can you speak to the testing that you've done with this code, and how one might independently verify the code on Baxter or another 7-DOF robot?

rethink-forrest:
I rebased off of master so that this commit only includes the changes to chainiksolvervel_pinv_nso.  The other changes are included in pull request #39 

I tested this on the Baxter 7-DOF arms for a Cartesian Path planning algorithm. Before these changes, we had no control over the end joint configuration - only the end Cartesian pose of the end effector. With these changes, we successfully biased the end joint configuration to a desired joint position. This was tested over a variety of straight Cartesian paths covering the robot's work space.

smits:
LGTM. Closes #46 
regarding #39, I'm would like to accept given we add the required docs to the header

PR #52, opened 2015-04-29
cmake: make Python version configurable

morxa:
On systems with both Python2 and Python3 available, one may want to
specify the Python version to use.


smits:
lgtm

PR #55, opened 2015-05-12
Added check in IK solver, chainiksolvervel_pinv_nso, for division by zero.

rethink-forrest:
Added a check that the normalizing term A is not zero during the nullspace optimization step. See line 124.

### DIFF
--- a/orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
+++ b/orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
@@ -90,11 +90,13 @@ namespace KDL
         // S^-1
         for (i = 0; i < nj; ++i) {
             Sinv(i) = fabs(S(i))<eps ? 0.0 : 1.0/S(i);
+        }
+        for (i = 0; i < 6; ++i) {
             tmp(i) = v_in(i);
         }
 
-        qdot_out.data = V * Sinv.asDiagonal() * U.transpose() * tmp;
-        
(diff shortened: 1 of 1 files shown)


rethink-forrest:
It's unclear why one of the four builds on Travis CI failed the tests. The tests passed on my computer, and the tests don't actually touch the chainiksolvervel_pinv_nso solver.

I tested this on the Baxter robot over 60 user defined trajectories covering the complete workspace.

rethink-forrest:
There was an instance of trying to access one element beyond the Twist vector which was only apparent when built with Asserts enabled.

The merge now passes the Travis CI tests.

orocos_kdl/src/chainiksolvervel_pinv_nso.cpp
```diff
@@ -121,18 +123,20 @@ namespace KDL
             A += qd*qd * weights(i)*weights(i);
         }
 
-        // Calculate inverse Jacobian Jc^-1
-        for (i = 0; i < nj; ++i) {
-            tmp(i) = weights(i)*(q_in(i) - opt_pos(i)) / A;
-        }
+        if (A > 1e-9) {
```
smits: I know we have more magic numbers like this in KDL, but could you elaborate on this one? Why 1e-9 and not 1e-12 like in other places?

rethink-forrest: As with most magic numbers, I don't have a strong reason, but 1e-9 seemed to work well during initial testing. At that value, the difference from the goal joint configuration, must be very small.

PR #58, opened 2015-07-15
Add possibility to return all the FK frames in a vector

ahoarau:
Just saving intermediate results.
It's like a robot state publisher in a way.

### DIFF
--- a/orocos_kdl/src/chainfksolverpos_recursive.cpp
+++ b/orocos_kdl/src/chainfksolverpos_recursive.cpp
@@ -63,8 +63,6 @@ namespace KDL {
         else
             segmentNr = seg_nr;
 
-        std::fill(p_out.begin(),p_out.end(),Frame::Identity());
-
         if(q_in.rows()!=chain.getNrOfJoints())
             return -1;
         else if(segmentNr>chain.getNrOfSegments())
@@ -72,6 +70,7 @@ namespace KDL {
         else if(p_out.size() != chain.getNrOfSegments())
             return -1;
         else{
+            std::fill(p_out.begin(),p_out.end(),Frame::Identity());
             int j=0;
             for(unsigned int i=0;i<segmentNr;i++){
                 if(chain.getSegment(i).getJoint().getType()!=Joint::None){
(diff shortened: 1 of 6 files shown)


orocos_kdl/src/chainfksolverpos_recursive.cpp
```diff
@@ -56,7 +56,34 @@ namespace KDL {
             return 0;
         }
     }
+    int ChainFkSolverPos_recursive::JntToCart(const JntArray& q_in, std::vector<Frame>& p_out, int seg_nr)    {
+        unsigned int segmentNr;
```
smits: You probably want to check that the p_out is of correct size!

orocos_kdl/src/chainfksolverpos_recursive.cpp
```diff
@@ -56,7 +56,34 @@ namespace KDL {
...
+        if(seg_nr<0)
+            segmentNr=chain.getNrOfSegments();
+        else
+            segmentNr = seg_nr;
+
+        std::fill(p_out.begin(),p_out.end(),Frame::Identity());
```
smits: Could you move the fill after line 74?

orocos_kdl/src/chainfksolvervel_recursive.cpp
```diff
@@ -63,5 +63,38 @@ namespace KDL
...
+    
+    int ChainFkSolverVel_recursive::JntToCart(const JntArrayVel& in,std::vector<FrameVel>& out,int seg_nr)
+    {
+        unsigned int segmentNr;
+        if(seg_nr<0)
+            segmentNr=chain.getNrOfSegments();
+        else
+            segmentNr = seg_nr;
+
+        std::fill(out.begin(),out.end(),FrameVel::Identity());
```
smits: Same here, move it after line 83

smits:
minor comments, lgtm otherwise

jf---:
Cool contribution...

ahoarau:
The fill is moved. 

Do you think the size of the vector the total number of segments or just segmentsNr ?

https://github.com/ahoarau/orocos_kinematics_dynamics/blob/fk_solvers_all_seg_frames_ret/orocos_kdl/src/chainfksolverpos_recursive.cpp#L70

smits:
I would say the size should be just segmentsNr.

ahoarau:
Ok it's done, but no idea why travis failed on that last build.

PR #61, opened 2015-11-09
Fix Fk Solvers return in vector 

ahoarau:
@smits I just realized my last PR https://github.com/orocos/orocos_kinematics_dynamics/pull/58 was wrong.
I fixed it and added a test for it. Sorry about that.


smits:
Thanks for the fixing and even more for the test.

PR #63, opened 2015-11-18
Feature/solvers use reference of chain

smits:
Requires #62 
- Adds the possibility to change a segment after adding it to a chain, like frame_to_tip and inertia, etc.
- Adds the possibility to change a chain after creating a solver for it. This requires some checking if the solvers internal data is still up-to-date with the number of joints of the chain.  Internal data is not resized under the hood to keep. A separate function updateInternalDatastructures is provided through SolverI to accomodate this. Memory-allocation are explicit in this case.


disRecord:
This change is a really bad design decision. 

First, reference to external object inside solver breaks incapsulations. Moreover, programmer have to manage memory in semi-manually manner and always keep in mind that `Chain` object **must outlive** `Solver` **without any address change**. So two object are so close coupled that they should be considered as one object.

Second, change of kinematic chain parameters can break solver if it contains internal structures which speed up calculations. For example, how should solver react if additional Segment is added to chain? Should it check chain size and possibly reconfigure itself in each method call? In this case reconfiguration is equal to recreation of the object.

Desired functionality can be achieved in different manner:
1. Pass  `Chain` as parameter to `JntToCart` and `CartToJnt`. 
2. Add setter for `Chain` or for chain's `Segment` in `Solver` class interface. Or simply leave this field `public` as it was in 1.3 KDL version.
3. Pass `shared_ptr` to constructor. This still is bad solution, due to second argument. 

First and third variants break API. But this kind of change **should break API**. Otherwise debug hell awaits all library users.

ahoarau:
I'm having an issue with this commit. I recently switch from `16.04 + kinetic + KDL 1.3.1` to an `18.04 + melodic + KDL 1.4` and it's **corrupting memory** somewhere.
Build melodic **from source** with KDL 1.3.1 fixed my issue.
Did you get anything similar @disRecord ?

disRecord:
Yes, we had the same issue.  Moreover it happened silently without  SEGFAULT signal so it was quite hard to debug it.

If you want to use KDL 1.4 your have to change your code. You should not destroy Chain object after Solver creation. 
For example, this commit https://gitlab.com/sweetie-bot/sweetie_bot_rt_control/commit/809c5bbc391a8678bef815356c1ed6ddaceb7322

ahoarau:
Thanks for the heads up.
Now I'm really wondering how do you copy a chain with this new API.

PR #70, opened 2016-01-25
Make orocos_kdl installspace relocatable.

mikepurvis:
Making the include paths relative to the find module makes it possible to move the installspace around after creation, including building it in conjunction with packages which depend on it, where the DESTDIR variable is set.


mikepurvis:
Hm. Looks like the travis job is failing due to an infrastructure problem, but hopefully this can be tested and merged at some point.

smits:
I'm fixing the travis job, I'll retrigger all pending PR builds once it's working again.

PR #64, opened 2015-11-23
Jacobian Time Derivative (Jdot) Solver 

ahoarau:
This PR adds the chainjnttojacdot solver that computes **Jdot**, a useful feature for calculating the cartesian acceleration : 
```
Xdot = J.qdot
Xddot = Jdot.qdot + J.qddot
```
It's an implementation of this paper : 
```
Symbolic differentiation of the velocity mapping for a serial kinematic chain
H. Bruyninckx, J. De Schutter
```
> http://www.sciencedirect.com/science/article/pii/0094114X9500069B

It supports the 3 representations : 
* **Body-fixed** : The reference frame is the end effector frame, and its origin is the velocity reference point.
* **Inertial** : The base is the reference frame, and its origin is the instantaneous velocity reference point.
* **Hybrid (*default*)**: The base is the reference frame, and the origin of the end effector frame is the velocity reference point (this is the default representation in KDL for the Jacobian).

Just call : 
```cpp
ChainJntToJacDotSolver jdot_solver(chain);

jdot_solver.setRepresentation( ChainJntToJacDotSolver::HYBRID ); //default
jdot_solver.setRepresentation( ChainJntToJacDotSolver::BODYFIXED );
jdot_solver.setRepresentation( ChainJntToJacDotSolver::INERTIAL );
// Hybrid is the default, so there's no need to call it 
jdot.JntToJacDot(q_in, jdot_out);

```

I also added a few tests : 
- A simple 2DOF arm for which I compute Jdot and compare it with the symbolic Jdot, for random q and qdot (in the hybrid representation) : error is <1e-10.
- 3 others arms for which I compute Jdot in the 3 representations, and compare it with [ J(q + qdot*dt) - J(q) ] / dt (which is , btw, not [that bad] (https://github.com/ahoarau/orocos_kinematics_dynamics/blob/feature/jdot_solver/orocos_kdl/tests/jacobiandottest.cpp#L255) :))

>> Featherstone's methods are (way) more efficient, but for a small chain (2x7dof) it's still pretty fast.

PR #76, opened 2016-02-10
Jacobian Time Derivative (Jdot) Solver: Follow-up

wannesvanloock:
This is a follow-up on #64 addressing the PR comments with some other minor fixes.

PTAL, @smits 
FYI, @ahoarau


ahoarau:
LGTM :+1:

PR #81, opened 2016-06-27
support for Visual Studio 2015 C++ compiler

mvistein:
This additional header was required to allow compilation with Visual Studio 2015 on Windows 7


smits:
lgtm

PR #83, opened 2016-07-06
Cherry picked fixes for 1.3

smits:


smits:
This should only contain binary compatible fixes for release-1.3.

PR #94, opened 2017-02-14
travis: new build matrix that include osx builds

smits:
Update to newer travis configuration api and include osx builds


.travis.yml
```diff
@@ -1,28 +1,102 @@
...
+    packages:
+      - libcppunit-dev
+      - libeigen3-dev
+      - python-future
+      - python-psutil
+      - python-sip-dev
+      - python3-future
+      - python3-psutil
+      - python3-sip-dev
+      - sip-dev
```
MatthijsBurgh, line 22: Do we need `sip-dev` if we have `python(3)-sip-dev`?

.travis.yml
```diff
@@ -1,28 +1,102 @@
 language: cpp
-os: linux
+os:
+  - linux
+  - osx
 dist: xenial
```
MatthijsBurgh, line 5: This isn't valid anymore

.travis.yml
```diff
@@ -1,28 +1,102 @@
...
   # build orocos_kdl
   - cd orocos_kdl
   - mkdir build
   - cd build
   - cmake -DENABLE_TESTS:BOOL=ON -DCMAKE_CXX_FLAGS:STRING="-Wall" -DCMAKE_BUILD_TYPE=${OROCOS_KDL_BUILD_TYPE} ./..
   # compile and install orocos_kdl
-  - make
-  - sudo make install
+  - make VERBOSE=1
+  - make install
```
MatthijsBurgh, line 99:
```suggestion
  - sudo make install
```
sudo is required in linux builds https://travis-ci.org/github/orocos/orocos_kinematics_dynamics/jobs/695656703#L646

.travis.yml
```diff
@@ -1,28 +1,102 @@
...
+      - libeigen3-dev
+      - python-future
+      - python-psutil
+      - python-sip-dev
+      - python3-future
+      - python3-psutil
+      - python3-sip-dev
+      - sip-dev
+
+matrix:
```
MatthijsBurgh, line 24:
```suggestion
jobs:
```

.travis.yml
```diff
@@ -1,28 +1,102 @@
 language: cpp
-os: linux
+os:
+  - linux
+  - osx
```
MatthijsBurgh, line 4:
```suggestion
os: linux
```

.travis.yml
```diff
@@ -1,28 +1,102 @@
...
 dist: xenial
 
 compiler:
```
MatthijsBurgh, line 7:
```suggestion
```

.travis.yml
```diff
@@ -1,28 +1,102 @@
...
   - gcc
```
MatthijsBurgh, line 8:
```suggestion
```

.travis.yml
meyerj: Not sure. The package contains the `/usr/bin/sip` executable, independent of `python-sip-dev` and `python3-sip-dev`. I think the CMake module [FindSIP.cmake](https://github.com/orocos/orocos_kinematics_dynamics/blob/984c9d0574d84f0a29a12471606aa736f14eb177/python_orocos_kdl/cmake/FindSIP.cmake#L16) requires the executable and exports it in `SIP_EXECUTABLE`.

meyerj:
@MatthijsBurgh Thanks for your input, but it does not make sense to work on this branch in parallel. For now I only wanted to merge the latest changes from branch `master` and resolve the obvious merge conflicts, and it was still work in progress. There are definitely more changes needed, e.g. to exclude Python 2 on macOS, and likely this requires to move the `before_script` and `script` sections into a separate all-in-one shell script with conditional sections.

> sudo is required in linux builds https://travis-ci.org/github/orocos/orocos_kinematics_dynamics/jobs/695656703#L646

`sudo` is not required because this patch originally changed the `CMAKE_INSTALL_PREFIX` to `$TRAVIS_BUILD_DIR/install`. But probably both approaches are valid. As far as I know `sudo` is not available for containerized builds.

I'll stop for today. Feel free to address the remaining issues from here on if you have time, or ping for help.

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: cpp
+os: linux
+dist: bionic
```
meyerj: `os` and `dist` are set per build job in the matrix configuration below. Apparently those two lines are redundant here and can be removed.

MatthijsBurgh: True, but Travis keeps throwing warnings if no root level `os` and `dist` are defined.

MatthijsBurgh:
@meyerj I think we should limit the number of jobs. Especially the OSX builds as these take so much more time.
I suggest we only run the release builds for OSX. (Maybe only one xcode version)
In Linux we should run the same set as we used to run, but in Bionic instead of Xenial. And only run the release builds also in Xenial.

Plus we could drop PyKDL based on SIP, which would also half the number of jobs.

MatthijsBurgh:
Not using travis anymore.

PR #106, opened 2017-09-05
kdl: added get sigma function to wdls solver

craigirobot:
Adding a function to read the singular values, sigma, so that the manipulability can be calculated from the product.

### DIFF
--- a/orocos_kdl/src/chainiksolvervel_wdls.cpp
+++ b/orocos_kdl/src/chainiksolvervel_wdls.cpp
@@ -85,9 +85,12 @@ namespace KDL
         maxiter=maxiter_in;
     }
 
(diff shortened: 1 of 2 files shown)


orocos_kdl/src/chainiksolvervel_wdls.cpp
```diff
@@ -85,6 +85,11 @@ namespace KDL
         maxiter=maxiter_in;
     }
 
+    void ChainIkSolverVel_wdls::getSigma(Eigen::VectorXd& Sout)
+    {
+        Sout=S;
```
smits: Shouldn't we also add a `E_SIZE_MISMATCH` check her like in the other functions? Otherwise this might lead to implicit resizing which we try to avoid.
What is the main driver to not just return a `const Eigen::VectorXd&`?

PR #107, opened 2017-09-26
avoiding printing status messages to stderr

dirk-thomas:
`message("some text")` is being printed to `stderr`. For status messages which are not warnings or errors that is not appropriate. Tgey should go to `stdout` instead.


dirk-thomas:
@smits Friendly ping.

PR #109, opened 2017-10-02
Fix some singularities when using Rotation::GetRotAngle()

daenny:
Using:
```python
import PyKDL as kdl

r = kdl.Rotation.RPY(3.14179265359, 1.57099632679, 0.0002)
print r
print "-"*30

axis_angle = r.GetRot()
a = axis_angle.Norm()
v = axis_angle / a

r2 = kdl.Rotation.Rot(v, a)
print r2
```
Results in two different rotations matrices. 
```
r =[     -0.0002,  3.7932e-12,          -1;
       -4e-08,          -1,  4.2068e-12;
           -1,       4e-08,      0.0002]
r2 = [     -0.0002,  2.5807e-08,           1;
   2.5807e-08,          -1, 2.58121e-08;
            1, 2.58121e-08,      0.0002]
```

similarly, `kdl.Rotation.RPY(-0.0002, 1.57059632679, 3.14139265359)`  Does not work. 

The new implementation is taken from:
http://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToAngle/index.htm

### DIFF
--- a/orocos_kdl/CMakeLists.txt
+++ b/orocos_kdl/CMakeLists.txt
@@ -74,7 +74,7 @@ endif(KDL_USE_NEW_TREE_INTERFACE)
 
 INCLUDE (${PROJ_SOURCE_DIR}/config/DependentOption.cmake)
 
-OPTION(ENABLE_TESTS OFF "Enable building of tests")
+OPTION(ENABLE_TESTS ON "Enable building of tests")
 IF( ENABLE_TESTS )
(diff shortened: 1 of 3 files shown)


daenny:
I had to change two tests, since the resulting vector is the other way around with this solution. 
Both solutions are identical though, since with 180 degrees rotation it does not matter which way the vector points.

PR #116, opened 2017-12-04
GetRot still seems to return NaNs for rotation matrices that contain a small error (not 100% orthogonal)

daenny:
We came across that the new implementation of "GetRot" still returns NaN, if the rotatation matrix is slightly non orthogonal. While this is of course not a valid rotation matrix, a small margin of error should be allowed. 

```
import PyKDL as kdl

f1 = kdl.Frame(kdl.Rotation(0.00923097,    0.969766,   -0.243866, 0.999943,  -0.0102479, -0.00290165, -0.00531284,   -0.243825,   -0.969805), kdl.Vector(0.112752,     1.26141,    0.250001))

f2 = kdl.Frame(kdl.Rotation(0.00924127,    0.969765,   -0.243866, 0.999943,  -0.0102586,  -0.0029021, -0.00531613,   -0.243825,   -0.969805), kdl.Vector(0.112757,     1.26141,   -0.298506))

def orthogonalize(f):
    R, P, Y = f.M.GetRPY()
    return kdl.Frame(kdl.Rotation.RPY(R, P, Y), kdl.Vector(f.p.x(), f.p.y(), f.p.z()))

print kdl.diff(f1, f2)
print kdl.diff(orthogonalize(f1), orthogonalize(f2))
```

### DIFF
--- a/orocos_kdl/src/frames.cpp
+++ b/orocos_kdl/src/frames.cpp
@@ -29,6 +29,7 @@
 
 #define _USE_MATH_DEFINES  // For MSVC
 #include <math.h>
+#include <algorithm>
 
 namespace KDL {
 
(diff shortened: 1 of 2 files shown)


smits:
Merged after manually resolving the conflicts in https://github.com/orocos/orocos_kinematics_dynamics/commit/fdc88f0b8

PR #118, opened 2017-12-12
Add Frame2 const operator*

mcopejans:


mcopejans:
@smits @meyerj PTAL

meyerj:
My best guess is that this change is fully ABI-compatible and there is no need to deprecate the non-const variant like in #119 because you are changing the const qualifier of a method declared inline. Compilers are not required to respect that "recommendation" and could have added a definition of `Vector2::operator*(const Vector2 &)` as a weak symbol to compilation units. But they cannot decide to not inline *and* to not generate code for the function and purely rely on external linkage. So there will always be a definition of whatever version requested available.

However, the only platform that I am aware of where we have automated builds is Linux/gcc. In this case the object files do not contain the `Vector2::operator*(const Vector2 &)` symbol and it should be safe to simply change the declaration and not keep the non-const variant. Users who compile from source most likely either do not update the source version or recompile all depending code afterwards, so no significant risk to break something.

Besides potential ABI-compatibility issues, the patch is perfectly valid.

mcopejans:
@meyerj Thanks for your response. The reasoning seems correct to me, so I will close the second PR.

PR #119, opened 2017-12-12
Add Frame2 const operator* and deprecate non const operator

mcopejans:


mcopejans:
Closing in favor of https://github.com/orocos/orocos_kinematics_dynamics/pull/118

PR #122, opened 2017-12-27
Fixed Vector2::Norm()

pdima:
It returned negative norm on vectors with negative x/y
and nan on Vector2(0,0)

### DIFF
--- a/orocos_kdl/src/frames.cpp
+++ b/orocos_kdl/src/frames.cpp
@@ -87,10 +87,16 @@ namespace KDL {
 
     double Vector2::Norm() const
     {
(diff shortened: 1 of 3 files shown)


orocos_kdl/src/frames.cpp
```diff
@@ -87,10 +87,16 @@ namespace KDL {
 
     double Vector2::Norm() const
     {
-        if (fabs(data[0]) > fabs(data[1]) ) {
```
smits: Why not leaving the code below and add `if (data[0] == 0 && data[1]==0) return 0.0;` as first statement of the function?

PR #125, opened 2018-01-09
Fixed __repr__ returns non string in Python 3

zchen24:
When built with Python 3, print Vector/Rotation/Frame returns repr returns non string error.
Fixed by replacing const char* with const std::string*. The fix has been tested to be backward compatible with Python 2


smits:
lgtm, thanks for the fix

Issue #123, opened 2018-01-09
framestest.py TabError: inconsistent use of tabs and spaces in indentation 


zchen24:
File "../tests/framestest.py", line 50
    self.assertEqual(t*2-t,t)
                            ^
TabError: inconsistent use of tabs and spaces in indentation

zchen24:
Merged. See #126

PR #126, opened 2018-01-10
Replaced tab with spaces in framestest.py and frameveltest.py

zchen24:
tabs and spaces are mixed in these two files, python throws Error. See issue #123


smits:
Thanks for cleaning that up.

PR #128, opened 2018-02-07
Fixed DeprecationWarning: Please use assertTrue instead. in PyKDL tests

zchen24:


meyerj:
LGTM.
For reference: https://bugs.python.org/issue9424

PR #132, opened 2018-02-20
Added q_min and q_max size mismatch check in ChainIkSolverPos_NR_JL

zchen24:
ChainIkSolverPos_NR_JL does not currently check q_min and q_max size. 
Added the check in the CartToJnt function.
I am also wondering if there is a good way to check this in the constructor.


smits:
👍

Issue #115, opened 2017-11-14
does KDL have python3 binding?


foolyc:
In python3, how to use KDL? or any other recommended tools?

ahcorde:
If you compile python_orocos_kdl with `catkin` using these flags: `-DPYTHON_VERSION=3.6 -DPYTHON_EXECUTABLE=/usr/bin/python3.5` you should be able to run it with Python3.

foolyc:
@ahcorde I complie it with py3.5, but when import the module PyKDL, segmentfault will appear.

Then I wrapper it with boost.Python [https://github.com/foolyc/PyKDL](url)

jacknlliu:
I think it needs a standalone release package for python3.

zchen24:
PyKDL is compatible with Python 3. Tested with Anaconda Python 3.6 version. 

![image](https://user-images.githubusercontent.com/2781007/35928320-c0134092-0be1-11e8-80e0-7bbb069d8d83.png)

To build: 
![image](https://user-images.githubusercontent.com/2781007/35928250-8c7dcb76-0be1-11e8-8f57-ce1ec9e9f36d.png)

smits:
@foolyc I cannot reproduce your problem, building the python bindings as @ahcorde suggests works for me, also when using python 3.5.  For the others I wonder how they got everything working without 74c3b80fe6ae

zchen24:
@MichaelLutter Can you try the master branch again? Make sure you have a clean build directory for python_orocos_kdl. (e.g. remove your existing build folder and create a new one)

MichaelLutter:
@zchen24  Thanks for the fast answer. When I try to compile *6fe130a*, I get compilation errors. I copied the compilation errors below.

Regarding the *Master Branch*, this branched has not changed recently. However, I deleted the build folder and built PyKDL from scratch and it still fails with 'updateInternalDataStructures'. 

"The compiled version is for Python 2, which does not have this PyInit_PyKDL function." What do you mean by that? Can one make PyKDL work under Python3? 


 *6fe130a Compilation Errors:* 
/.../orocos_kinematics_dynamics/python_orocos_kdl/build/PyKDL/sipPyKDLpart2.cpp: In function ‘PyObject* meth_ChainIkSolverVel_pinv_nso_getWeights(PyObject*, PyObject*)’:
[...]

PR #145, opened 2018-05-14
Fixed python3 std string conversion issue

zchen24:
@smits Seems to fix some string conversion issue in Python 3. Please also see #115 
@MichaelLutter this should fix the bug.


MichaelLutter:
@zchen24 Thats awesome. Thanks a lot for the fast Bug Fix! For me this Pull Request ist working under Python3 and MacOS.

PR #151, opened 2018-06-09
Fixed velocityprofile_dirac does not stop bug and replaced tab with 4…

zchen24:


smits:
Looks good, it would be nice if you could add a unittest for this.

zchen24:
@smits Thanks for reviewing. A unit test has been added.

zchen24:
@smits Could you please take a look at the added unit-test? Thanks.

orocos_kdl/tests/velocityprofiletest.cpp
```diff
@@ -334,3 +334,45 @@ void VelocityProfileTest::TestTrapHalf_SetDuration_End()
...
+    v.SetProfile(pos1, pos2);
+
+    time = 0;
+    CPPUNIT_ASSERT_DOUBLES_EQUAL(pos1, v.Pos(time), epsilon);
+
+    time = 1.0;
+    CPPUNIT_ASSERT_DOUBLES_EQUAL(pos2, v.Pos(time), epsilon);
+
+    time = -1.0;
+    CPPUNIT_ASSERT_DOUBLES_EQUAL(pos2, v.Pos(time), epsilon);
```
smits: Shouldn't this be `pos1`?

zchen24: @smits Updated the source code and unit-test.

PR #152, opened 2018-06-15
Added ChainIdSolver python wrapper

zchen24:
The Vector.sip snippet is from https://github.com/ifwe/wxpy/blob/master/src/vector.sip


jbohren:
Thanks, Z.

This works as advertised.

PR #154, opened 2018-06-27
Adding catkin env-hooks to support Python venv

jbohren-hbr:
PyKDL gets installed into `PYTHON_SITE_PACKAGES_INSTALL_DIR`, which doesn't end up in the `PYTHONPATH` by default and fails when build in a python virtual environment.


meyerj:
Symlinking for all various shells is not necessary. catkin uses `sh` as a catch-all shell, so only [python_orocos_kdl_site_packages.sh.in](https://github.com/orocos/orocos_kinematics_dynamics/pull/154/files#diff-30f736618a4529f071efc58166eff268) and the following line in [CMakeLists.txt](https://github.com/orocos/orocos_kinematics_dynamics/pull/154/files#diff-1b9c0761d710b10cf0daf317fb291b5f) would be sufficient:

```cmake
catkin_add_env_hooks(python_orocos_kdl_site_packages SHELLS sh DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/env-hooks)
```

It's poorly documented. Check https://github.com/ros/catkin/blob/kinetic-devel/cmake/templates/_setup_util.py.in#L198.

PR #159, opened 2018-07-07
Change auto_ptr to shared_ptr, since auto_ptr is deprecated.

allenh1:
Trivial change -- just updates the code a bit.

---
This change is [Reviewable](https://reviewable.io/reviews/orocos/orocos_kinematics_dynamics/159)

### DIFF
--- a/orocos_kdl/src/path_circle.cpp
+++ b/orocos_kdl/src/path_circle.cpp
@@ -55,7 +55,9 @@ Path_Circle::Path_Circle(const Frame& F_base_start,
 			RotationalInterpolation* _orient,
 			double _eqradius,
             bool _aggregate) :
-				Path_Circle(_orient, _eqradius, _aggregate)
+				orient(_orient) ,
+				eqradius(_eqradius),
+                aggregate(_aggregate)
 			{
 					F_base_center.p = _V_base_center;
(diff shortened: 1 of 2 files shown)


meyerj:
In order to stay backwards-compatible to pre-C++11 compilers, this patch should add preprocessor checks whether `std::shared_ptr` is available, e.g. by adding
```cpp
#if __cplusplus < 201103L
  #define shared_ptr auto_ptr
#endif
```

Or using `boost::shared_ptr` instead:
```cpp
#if __cplusplus < 201103L
  #include <boost/shared_ptr.hpp>
  using boost::shared_ptr;
#endif
```

But actually in most, if not in all cases, a [unique_ptr](http://www.cplusplus.com/reference/memory/unique_ptr/) will do instead of a `shared_ptr` and is slightly more efficient.

allenh1:
> In order to stay backwards-compatible to pre-C++11 compilers, this patch should add preprocessor checks whether std::shared_ptr is available

That sounds reasonable to me. 

> But actually in most, if not in all cases, a unique_ptr will do instead of a shared_ptr and is slightly more efficient.

Yes, this is also true. I think I was just being baraded with patches to throw at people and decided shared_ptr for whatever reason. `std::unique_ptr` should be a drop in for `std::auto_ptr`.

I'll update the PR when I have some time.

PR #165, opened 2018-08-20
fix INSTALL instructions

qqfly:
fix bug in INSTALL: ccmake --> cmake


meyerj:
Not strictly a bug. `ccmake` is the [Curses Dialog interface](https://cmake.org/cmake/help/latest/manual/ccmake.1.html) of CMake, but some users might not have it installed.

PR #167, opened 2018-09-21
Remove const reference'd chains from solvers. Fixes #147

### LINKED ISSUE #147: Should ChainIkSolverPos__LMA store a KDL::Chain as a const ref?
When the Levenberg-Marquardt solver stores the kinematic chain as a const ref,

https://github.com/orocos/orocos_kinematics_dynamics/blob/a7a8d282a5a94948176f2c5af58fbd047d849030/orocos_kdl/src/chainiksolverpos_lma.hpp#L160

It allows scenarios where the chain passed in the constructor is destroyed, and the solver will eventually segfault when its called, without warning for what has happened. If the intention is to not copy the chain, a shared pointer should be used. 

However, the other solvers seem to be storing a copy of the chain, and don't have this issue. 

Just spent a couple hours trying to understand why suddenly my IK solver was segfaulting :)

diogoalmeida:
Storing the chains as const references is risky, as the passed chain can be destroyed outside of the solver.

### DIFF
--- a/orocos_kdl/src/chaindynparam.hpp
+++ b/orocos_kdl/src/chaindynparam.hpp
@@ -58,7 +58,7 @@ namespace KDL {
     virtual void updateInternalDataStructures();
 
     private:
-        const Chain& chain;
+        const Chain chain;
 	int nr;
 	unsigned int nj;
         unsigned int ns;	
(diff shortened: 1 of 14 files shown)


smits:
Storing chains as references is intentional, see https://github.com/orocos/orocos_kinematics_dynamics/pull/63

PR #169, opened 2018-11-07
Adding the Jacobian Dot Solver to PyKDL

MichaelLutter:
PyKDL Changes:
- Exposing the Jacobian Dot Solver 
- Adding the optional Segment index to Jacobian- and Jacobian Dot Solver


smits:
Thanks for your contribution.

PR #171, opened 2018-12-29
Create chain forward dynamics solver class

craigirobot:
kdl solver: added chain forward dynamics solver

Created forward dynamics solver class based on inverse dynamics solver and inertia matrix from chaindynparam solver. In addition, a LDL solver function was added under the utilities to calculate the joint acceleration from the inertia matrix inverse times the joint torques.

### DIFF
--- a/orocos_kdl/src/chainfdsolver.hpp
+++ b/orocos_kdl/src/chainfdsolver.hpp
@@ -1,0 +1,61 @@
+// Copyright  (C)  2018  Ruben Smits <ruben dot smits at mech dot kuleuven dot be>
+
+// Version: 1.0
+// Author: Ruben Smits, Craig Carignan
+// Maintainer: Ruben Smits <ruben dot smits at mech dot kuleuven dot be>
+// URL: http://www.orocos.org/kdl
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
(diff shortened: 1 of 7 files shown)


DjoleMNE:
Hi Craig,
thank for developing the FD solver and RK4 integrator for it! 
Can you comment on the complexity of the developed algorithm and give a reference for its theoretical background?

Additionally, I wanted to point out that there already exists an FD algorithm in KDL. More specifically, (Popov-) Vereschagin constrained Hybrid Dynamics algorithm, which can be used for solving both inverse and forward dynamics problems. Here its is implemented under name "chainidsolver_vereshchagin", where id part in the name is a little bit misleading. 
Nevertheless, the solver has O(n) complexity and the robot task can be specified using 3 interfaces:
* End-effector Acceleration Constraint interface, by setting "alpha" and "beta" variables/arguments.
* External force interface, by setting "f_ext" variable/argument that describes external forces acting on each robot segment (note that these forces need to be expressed w.r.t. base frame)
* Feedforward torque interface, by setting "torques" variable/argument.

If a user wants to use the solver for only FD, it is necessary to define its joint torque values via the third interface and set values for the other two interfaces to zero. Of course, the gravity acceleration is defined separately in solver's constructor.

A user can see in https://github.com/orocos/orocos_kinematics_dynamics/blob/master/orocos_kdl/tests/solvertest.cpp file (function SolverTest::VereshchaginTest() ) an example on how to use the solver.

For more details about the solver, especially theoretical, a user can read chapters from following two PhD dissertations:
* Shakhimardanov, Azamat: Composable Robot Motion Stack: Implementing Constrained Hybrid Dynamics using Semantic Models of Kinematic Chains
https://limo.libis.be/primo-explore/fulldisplay?docid=LIRIAS1747300&context=L&vid=Lirias&search_scope=Lirias&tab=default_tab&lang=en_US&fromSitemap=1

* Smits, Ruben: Robot Skills: Design of a Constraint-Based Methodology and Software Support
https://limo.libis.be/primo-explore/fulldisplay?docid=LIRIAS1745778&context=L&vid=Lirias&search_scope=Lirias&tab=default_tab&lang=en_US&fromSitemap=1


I think here in KDL, more documentation and examples should exist for users to see how to use the solver for different problems, because it gives more features that the classical RNE, at least the ID version.

Best regards,
Djordje Vukcevic

DjoleMNE:
Hi @craigirobot,

sorry for the late update!

I made a pull request recently with documentation and improvements on interfaces, computations and test cases for the Vereshchagin solver. See #280 for more details.

However, I still haven't compared timings of these two solvers (Vereshchagin and your FD). But I am planning to do that as well soon. Just wanting for @smits and other guys to review the pull request before I add anything new.

Best regards,
Djordje

PR #172, opened 2019-01-09
Fix eigen alignment issues and replace auto_ptr with unique_ptr

jsqabcjsq:
1. Move all eigen types into eigen_types.hpp and set the alignment to be DontAlign.
    If we don't do this, we need to change the behavior of operator new for every class who has eigen types as member, or it will cause memory corruption issues. For details, please refer to: http://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html

2. Replace auto_ptr with unique_ptr. I noticed there's a pull request on doing the same job but with shared_ptr. Upgrading to C++11 is unavoidable, the better way for supporting legacy compilers, IMHO, could be suggesting users to checkout certain version.


meyerj:
> 1. Move all eigen types into eigen_types.hpp and set the alignment to be DontAlign.
>    If we don't do this, we need to change the behavior of operator new for every class who has eigen types as member, or it will cause memory corruption issues. For details, please refer to: http://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html

This issue [only refers to various fixed-size types](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html#title5) with a size that is a multiple of 16 bytes. Many of the matrices and vectors you replaced are dynamically sized and will be actually vectorized by Eigen. We definitely do not want to add `DontAlign` for dynamically-sized matrices, although I am not sure whether it would have an effect at all.

Adding `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` in classes where this is required is an option because not doing so would be a serious bug, but I could not find a place where one of the affected types is used other than on the stack. [`Eigen::Map`s are `Unaligned` by default.](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html)

I can imagine that some non-Intel platforms or new processor features like AVX512 have different alignment requirements that were not covered yet. Is this the case for you? Did you experience actual errors (or failed assertions in debug mode) when using KDL?


> 2. Replace auto_ptr with unique_ptr. I noticed there's a pull request on doing the same job but with shared_ptr. Upgrading to C++11 is unavoidable, the better way for supporting legacy compilers, IMHO, could be suggesting users to checkout certain version.

This topic has already been addressed in #187. For now we do not want to depend on C++11 yet, at least not without a bigger new release and more complete API overhaul.


Anyway, thanks for your contribution and sorry for the long delay since you posted it! Please close the issue if you agree with my conclusions, or comment if you still think the patch is necessary.

meyerj:
I am closing this issue for the lack of activity, and because in the current state the patch is invalid.
Feel free to reopen it or comment if needed.

PR #173, opened 2019-01-09
Extend Recursive Newton-Euler for Inverse Dynamics of Trees

francofusco:
Inverse Dynamics is supported in KDL for chains only. This pull request introduces the abstract solver `TreeIdSolver` to do inverse dynamics on a full tree. The concrete implementation `TreeIdSolver_RNE` is a trivial extension of the corresponding solver for chains. It uses `std::map`s rather than `std::vector`s to store kinematic/dynamic quantities for each segment.

The algorithm has been tested in two scenarios:

- A tree featuring two "independent" chains, _ie_, both starting at the root of the tree. The efforts returned by the solver are compared with the ones returned by two solvers operating on each chain separately.
- A tree with two leafs attached to a common segment, which is in turn attached to the root. The efforts are compared with the analytic model for validation.

Note: the first commit is the one introducing the new features, the latter two are just cleaning an fixing some mistakes I introduced.


smits:
Thanks for the contribution, it seems that the tests are still failing.  If you fix that we can merge it.

francofusco:
Sorry, I had some busy weeks and I did not check the status for a while. Everything should have been fixed now!

PR #174, opened 2019-01-09
New Tree features: sub-tree extraction and better tree printing

francofusco:
The two commits introduce one new feature each:

1. A new public method `Tree::getSubTree(const string& segment_name, Tree& tree)`. This allows to extract from the original object a new tree that starts at `segment_name` and includes all its children.

2. Two functions `tree2str` (see [kinfam_io.hpp](https://github.com/francofusco/orocos_kinematics_dynamics/blob/new_tree_features/orocos_kdl/src/kinfam_io.hpp)) which return a string representing a tree. In my opinion they provide a better and clearer output than `operator<<(ostream&, Tree&)` thanks to the use of indentation.

### DIFF
--- a/orocos_kdl/src/kinfam_io.cpp
+++ b/orocos_kdl/src/kinfam_io.cpp
@@ -21,6 +21,7 @@
 
 #include "kinfam_io.hpp"
 #include "frames_io.hpp"
+#include <string>
 
 namespace KDL {
 std::ostream& operator <<(std::ostream& os, const Joint& joint) {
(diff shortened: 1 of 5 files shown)


francofusco:
Sorry to bother, I just wanted to know if there is any reason why this PR is not being merged. If something is wrong with it, I can change it. I am particularly interested in having the `getSubTree` method added to KDL, mainly to be able to use the Inverse Dynamics solver on "smaller" trees extracted from a URDF.

francofusco:
My bad, sorry. Thanks for having found the time!

PR #176, opened 2019-01-28
[Windows] Fix SIPMacros.cmake for Windows build.

seanyen-msft:
### Background
SIP takes .sip to generate C++ code to build native Python binding module, and SIP_CONCAT_PARTS decides how many C++ files to generate for parallel build. And since it doesn't know how many parts SIP actually created in Cmake configuration time, it emits command to generate the stub C++ files in case they are not there.

### Problem
The "touch" implementation seems to be broken on Windows. It is generating a file called "filename" under the current directory, instead of touching and creating the stub files.

### Solution
Instead of special handling the touch commands for Windows, replace the logic with a more generic "COMMAND ${CMAKE_COMMAND} -E touch ${_sip_output_files}" to be cross platform.

(This is verified on **ROS on Windows** project. https://aka.ms/ros)


seanyen-msft:
Ping!

seanyen-msft:
@smits checking the commit history, since you were spending lots of valuable time on code review and merge for this repository. Hope I can get some feedbacks on this change. Thanks!

seanyen-msft:
Thanks!

PR #178, opened 2019-02-28
Add private constructor for delegating initialization. Fixes #177.

### LINKED ISSUE #177: Memory leak: Path_Circle does not delete orient ptr
When `Path_Circle()` throws an exception, initialization of the object is not completed. In this case the destructor is not called, which is responsible for `delete orient`.

This issue could be resolved by delegating object initialization.

martiniil:


smits:
@martiniil I did not notice this broke our build, please check https://travis-ci.org/orocos/orocos_kinematics_dynamics/jobs/499811973

martiniil:
@smits Thanks for your approval. Building with c++11 is not an option? Then I would have to think about another way.

smits:
It might be an option, but then we should enforce it in our build to avoid accidental breakage ;)

PR #180, opened 2019-03-22
Fix memory leak in Path_Circle constructor. Fixes #177

### LINKED ISSUE #177: Memory leak: Path_Circle does not delete orient ptr

martiniil:
This should work. I am open for discussion.

### DIFF
--- a/orocos_kdl/src/path_circle.cpp
+++ b/orocos_kdl/src/path_circle.cpp
@@ -59,34 +59,39 @@ Path_Circle::Path_Circle(const Frame& F_base_start,
 				eqradius(_eqradius),
                 aggregate(_aggregate)
 			{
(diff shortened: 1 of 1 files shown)


martiniil:
@smits The exceptions are thrown if the object cannot be constructed. I think this is a proper solution.
At least this change prevents the memory leak.

Of course, we could think of refactoring the whole class. But I am not aware of the consequences :)

MatthijsBurgh:
@martiniil @meyerj I think this is a more elegant solution.
`delete ptr, throw` instead of `throw, catch, delete ptr, throw`

PR #182, opened 2019-04-09
Replace auto_ptr with unique_ptr

allenh1:
A re-do of #159 (force pushed and it got closed automatically -- sorry for the noise).

### DIFF
--- a/orocos_kdl/src/path.cpp
+++ b/orocos_kdl/src/path.cpp
@@ -59,7 +59,7 @@ using namespace std;
 
 
 Path* Path::Read(istream& is) {
-	// auto_ptr because exception can be thrown !
+	// unique_ptr because exception can be thrown !
 	IOTrace("Path::Read");
 	char storage[64];
 	EatWord(is,"[",storage,sizeof(storage));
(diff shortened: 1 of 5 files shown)


smits:
AFIAU this PR got obsolete by #187 @francisco-miguel-almeida @allenh1 can you confirm? If so we can just close this PR.

allenh1:
> AFIAU this PR got obsolete by #187 @francisco-miguel-almeida @allenh1 can you confirm?

@smits Yes, that one seems to have made this obsolete.

But, if I may, why add a `scope_ptr` in place of `std::unique_ptr`?

meyerj:
> > AFIAU this PR got obsolete by #187 @francisco-miguel-almeida @allenh1 can you confirm?
> 
> @smits Yes, that one seems to have made this obsolete.
> 
> But, if I may, why add a `scope_ptr` in place of `std::unique_ptr`?

In case of #187 only compiled sources are affected, so a preprocessor conditional as suggested in https://github.com/orocos/orocos_kinematics_dynamics/pull/159#issuecomment-434132651 would have been an option to use `std::unique_ptr`, if available. But public API in headers should not use any C++1x features, at least not for now.

PR #197, opened 2019-10-24
Simplify implementation of addDelta() for rotations

meyerj:
`Rot(M*x) == M*Rot(x)*M^T` for any rotation matrix `M` and vector x, where `Rot(x)` returns the rotation matrix that corresponds to rotation vector `x` according to the [Rodrigues rotation formula](https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula), as implemented in [Rot()](https://github.com/orocos/orocos_kinematics_dynamics/blob/1ae45bb2e0821586e74047b01f7cb48d913204f4/orocos_kdl/src/frames.inl#L1095-L1121).

So the expression

    a*Rot(a.Inverse(da)*dt)

in `addDelta(const Rotation& a,const Vector&da, double dt)` can be simplified as

    = a*Rot(a.Inverse()*da*dt)
    = a*a.Inverse()*Rot(da*dt)*a
    = Rot(da*dt)*a

which saves one matrix multiplication.

### DIFF
--- a/orocos_kdl/src/frames.inl
+++ b/orocos_kdl/src/frames.inl
@@ -1151,7 +1151,7 @@ IMETHOD Vector addDelta(const Vector& a,const Vector&da,double dt) {
 }
 
 IMETHOD Rotation addDelta(const Rotation& a,const Vector&da,double dt) {
-	return a*Rot(a.Inverse(da)*dt);
+	return Rot(da*dt)*a;
 }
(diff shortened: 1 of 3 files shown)


meyerj:
> we do not have any unit-test protecting this function  🤔 

added in https://github.com/orocos/orocos_kinematics_dynamics/pull/197/commits/552c26cddf316382d1fcca1ba55226d9f0259ecf :heavy_check_mark:

PR #199, opened 2019-11-08
back to use a more numerically stable method to compute the rotation angle

maverick-long:
This PR is related to the recent 1.3.2 release.

When I used KDL to compute a very precise ik solution (eps = 1e-6 or 1e-7), the `angle = acos()` in `Rotation::GetRotAngle(Vector& axis,double eps)` can not provide a highly accurate rotation angle, which can cause ik fail. According to https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation, the previous `angle = atan2()` method is more numerically stable. 

After I switched to `angle=atan2()`, the succeed rate increased a lot.

### DIFF
--- a/orocos_kdl/src/frames.cpp
+++ b/orocos_kdl/src/frames.cpp
@@ -416,12 +416,12 @@ double Rotation::GetRotAngle(Vector& axis,double eps) const {
     // If the matrix is slightly non-orthogonal, `f` may be out of the (-1, +1) range.
     // Therefore, clamp it between those values to avoid NaNs.
     double f = (data[0] + data[4] + data[8] - 1) / 2;
-    angle = acos(std::max(-1.0, std::min(1.0, f)));
 
     x = (data[7] - data[5]);
     y = (data[2] - data[6]);
     z = (data[3] - data[1]);
     axis = KDL::Vector(x, y, z);
+    angle = atan2(axis.Norm()/2,f);
     axis.Normalize();
     return angle;
 }
--- a/orocos_kdl/src/frames.cpp
+++ b/orocos_kdl/src/frames.cpp
@@ -413,8 +413,6 @@ double Rotation::GetRotAngle(Vector& axis,double eps) const {
         return angle; // return 180 deg rotation
     }
 
-    // If the matrix is slightly non-orthogonal, `f` may be out of the (-1, +1) range.
-    // Therefore, clamp it between those values to avoid NaNs.
     double f = (data[0] + data[4] + data[8] - 1) / 2;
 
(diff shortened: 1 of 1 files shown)


orocos_kdl/src/frames.cpp
```diff
@@ -413,15 +413,13 @@ double Rotation::GetRotAngle(Vector& axis,double eps) const {
...
-    // If the matrix is slightly non-orthogonal, `f` may be out of the (-1, +1) range.
-    // Therefore, clamp it between those values to avoid NaNs.
     double f = (data[0] + data[4] + data[8] - 1) / 2;
-    angle = acos(std::max(-1.0, std::min(1.0, f)));
 
     x = (data[7] - data[5]);
     y = (data[2] - data[6]);
     z = (data[3] - data[1]);
     axis = KDL::Vector(x, y, z);
+    angle = atan2(axis.Norm()/2,f);
```
meyerj: Not directly related, but I think the tests for singularities above can be simplified significantly if they would be based on the pre-computed values for the general case
```cpp
double c = (data[0] + data[4] + data[8] - 1) / 2;    // == f == cos(angle)
```
and
```cpp
double s = axis.Norm()/2;    // == sin(angle)
```
Only the `atan(s,c)` and `axis.Normalize()` calls would fail for the corner cases `angle == 0` or `angle == 180`.

So
```cpp
    if ((std::abs(data[1] - data[3]) < epsilon)
        && (std::abs(data[2] - data[6])< epsilon)
        && (std::abs(data[5] - data[7]) < epsilon))
```
is equivalent to
```cpp
    if (s < epsilon)  // angle is close to 0 or 180 degrees
```
and the inner check
```cpp
        if ((std::abs(data[1] + data[3]) < epsilon2)
            && (std::abs(data[2] + data[6]) < epsilon2)
            && (std::abs(data[5] + data[7]) < epsilon2)
            && (std::abs(data[0] + data[4] + data[8]-3) < epsilon2))
```
is equivalent to
```cpp
    if ((std::abs(c - 1) < epsilon)  // cos(angle) is approximately 1 => angle is 0
```

Furthermore
```cpp
double half_sqrt_2 = 0.5 * sqrt(2.0);
```
is unused.

smits:
@meyerj I'd rather forward-port bugfixes than backport them.

meyerj:
@smits Should we release this patch into Kinetic? The latest version [1.3.2](https://github.com/orocos/orocos_kinematics_dynamics/releases/tag/1.3.2) still does not have this fix.

smits:
@meyerj sounds like a good plan to create a patch release for kinetic

PR #215, opened 2020-02-03
Fix build warnings in gcc 9.2.1

ivanpauno:
I got those warnings in the [ros2 fork](https://github.com/ros2/orocos_kinematics_dynamics).
Here an example of them:
https://ci.ros2.org/job/ci_linux/9200/warnings23Result/package.274912723/

### DIFF
--- a/orocos_kdl/src/frames.hpp
+++ b/orocos_kdl/src/frames.hpp
@@ -310,8 +310,9 @@ public:
                 double Xy,double Yy,double Zy,
                 double Xz,double Yz,double Zz);
     inline Rotation(const Vector& x,const Vector& y,const Vector& z);
-    // default copy constructor is sufficient
 
+    // default copy constructor is sufficient
+    constexpr Rotation(const Rotation&) = default;
 
      inline Rotation& operator=(const Rotation& arg);
 
@@ -1059,6 +1060,9 @@ public:
     explicit Rotation2(double angle_rad):s(sin(angle_rad)),c(cos(angle_rad)) {}
 
     Rotation2(double ca,double sa):s(sa),c(ca){}
+
+    // default copy constructor is sufficient
+    constexpr Rotation2(const Rotation2&) = default;
 
      inline Rotation2& operator=(const Rotation2& arg);
      inline Vector2 operator*(const Vector2& v) const;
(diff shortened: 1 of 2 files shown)


ivanpauno:
This warnings appear when building with Ubuntu 20.04.

meyerj:
If I understand [the description of `-Wdeprecated-copy`](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/C_002b_002b-Dialect-Options.html#index-Wdeprecated-copy) correctly, the problem is that `Rotation` and `Rotation2` have a user-defined assignment operator but no user-defined constructor. This is indeed a violation of the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)).

The implicitly defined default copy and assignment operators should be just fine and there is no user-defined destructor, so another option to fix the warning would be to simply remove the declaration of `operator=`. But that would break the ABI, and therefore your patch LGTM.

We should check whether there are other user-defined copy constructors or assignment operators that are equivalent to the default and remove them for the next major release, for example in class `Frame` (@smits?). Just an assumption, but the default ones might be faster than copying the rotation matrix element-by-element in a for-loop (equivalent to `memcpy`).

ivanpauno:
> If I understand the description of -Wdeprecated-copy correctly, the problem is that Rotation and Rotation2 have a user-defined assignment operator but no user-defined constructor. This is indeed a violation of the Rule of Three.

Yes!

> The implicitly defined default copy and assignment operators should be just fine and there is no user-defined destructor, so another option to fix the warning would be to simply remove the declaration of operator=. But that would break the ABI, and therefore your patch LGTM.

I'm not compltly sure if deleting the explicit defined assignment operator breaks ABI or not (neither if it's compiler dependent or not).

meyerj:
> I'm not compltly sure if deleting the explicit defined assignment operator breaks ABI or not (neither if it's compiler dependent or not).

It would. Binaries compiled against the old header expect to find the symbol of the assignment operator in the shared library, and the compiler did not define an implicit inline operator.

meyerj:
GCC 9.2.1 outputs many more of those warnings if `-Wdeprecated` is in place, primarily in Eigen headers, but also in other KDL declarations. Ideally they should also be fixed before the Noetic release (#218).

See [here](https://gist.github.com/meyerj/911c868042b385e281e5edfecc4f2492) for the full output (Eigen excluded).

Issue #211, opened 2020-01-30
Memory Leak in PyKDL Tree.getChain


mag-sruehl:
This code will consume will grow the memory usage on my pc from 107 mb at the first raw_input to 677 mb in the end:

```
import kdl_parser_py.urdf as urdf
import gc

robot = """<?xml version="1.0"?>
<robot name="visual">
  <link name="base_link"/>
  <link name="right_leg"/>
  <joint name="base_to_right_leg" type="fixed">
    <parent link="base_link"/>
    <child link="right_leg"/>
  </joint>
</robot>
"""
raw_input()

tree = urdf.treeFromString(robot)[1]
for i in range(1000000):
    tree.getChain("base_link", "right_leg")
gc.collect()
raw_input()
```

valgrind output (only the very last):
```
==14776== 545,924,946 (39,999,200 direct, 505,925,746 indirect) bytes in 999,980 blocks are definitely lost in loss record 712 of 712
==14776==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14776==    by 0xB66143F: meth_Tree_getChain (in /opt/ros/melodic/lib/python2.7/dist-packages/PyKDL.so)
==14776==    by 0x1FCE69: call_function (ceval.c:4376)
==14776==    by 0x1FCE69: PyEval_EvalFrameEx (ceval.c:3013)
==14776==    by 0x1FA619: PyEval_EvalCodeEx (ceval.c:3608)
==14776==    by 0x1F9F38: PyEval_EvalCode (ceval.c:669)
==14776==    by 0x22B04E: run_mod.lto_priv.2104 (pythonrun.c:1385)
==14776==    by 0x226291: PyRun_FileExFlags (pythonrun.c:1371)
==14776==    by 0x225CBC: PyRun_SimpleFileExFlags (pythonrun.c:957)
==14776==    by 0x1D43E5: Py_Main (main.c:645)
==14776==    by 0x4E5DB96: (below main) (libc-start.c:310)
```

I used  orocos_kinematics_dynamics 1.4.0 and current master. As this is similar to #157, I have built it with with the fixed 4.19-maint (v4.19.21 ??? January 2020)  sip version.

Further, I suspect the problem is the new kinfam.sip:271
```
    Chain* getChain(const std::string& chain_root, const std::string& chain_tip)const;
%MethodCode
    Chain* chain = new Chain();
[...]

PR #216, opened 2020-02-04
added a test for the memory leak in Tree.getChain from #211

mag-sruehl:
Checks that a sequence of getChain calls does not increase the memory usage.

I am not sure what the test dependency on python-psutil will break on some build systems, but it seems to be the recommended os independent way to get the memory usage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
 
 before_script:
   - sudo apt-get -qq update
-  - sudo apt-get install -y libeigen3-dev libcppunit-dev python-sip-dev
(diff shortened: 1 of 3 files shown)


mag-sruehl:
the travis before script would need to know about psutil:
sudo apt-get install -y libeigen3-dev libcppunit-dev python-sip-dev python-psutil

meyerj:
Thanks for your contribution and especially for digging into the leaking memory issue (#211)!
I added the additional python package to the Travis config in 74c72a6 and now Travis CI is happy.

Issue #218, opened 2020-02-07
Noetic release?


sloretz:
It looks like all dependencies of packages in this repo have been released into Noetic. Mind releasing this repo? That would unblock releasing repos which depend on this one.

http://repositories.ros.org/status_page/blocked_releases_noetic.html?q=id%3D%22orocos_kinematics_dynamics%22&s=6&r=1

smits:
Ok, I'll talk to some people here to get it released.

sloretz:
Might not need a Noetic release if https://github.com/ros/rosdistro/pull/23841 is merged

sloretz:
It looks like this is no longer necessary. ROS Noetic will use the releases in the upstream Focal and Buster apt repos. Thanks for your time!

PR #222, opened 2020-02-19
Conditional dependencies

sloretz:
This makes the `python_orocos_kdl` package use conditional dependencies and `ROS_PYTHON_VERSION` for `python-sip` and `python3-sip`. This is necessary to use rosdep to install dependencies for this package since the `master` branch will support both Python 2 and Python 3.  It was also required to delete the `manifest.xml` since it has an unconditional dependency on the rosdep key `python-sip`, which cannot be resolved on Ubuntu Focal.

See also http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_dependencies_in_your_package.xml

required for #218


python_orocos_kdl/manifest.xml
```diff
@@ -1,22 +0,0 @@
...
-  <rosdep name="python-sip"/>
-
-  <export>
-    <python path="${prefix}/lib"  />
-    <rosdoc config="rosdoc.yaml"/>
-  </export>
-
-</package>
-
-
```
meyerj: Be aware that `manifest.xml` is not only used by older ROS versions, but also by other build tools used by the Orocos community, like [Autoproj](https://github.com/rock-core/autoproj). We also parse it in [rtt/UseOROCOS-RTT-helpers.cmake](https://github.com/orocos-toolchain/rtt/blob/49752330ac2290ffb5206d42245ace716e2d93db/UseOROCOS-RTT-helpers.cmake#L54-L93). The two specs used to be compatible a long time ago, but also Autoproj added new features that were not understood by the ROS tools.

The dual use of `manifest.xml` already caused some trouble in the past. I once proposed to prefer `package.xml` if it exists (https://github.com/ros-infrastructure/rosdep/issues/353), but that was rejected back then. I noticed that the issue surfaced again recently (https://github.com/ros-infrastructure/rosdep/pull/737).

In this particular case it should be fine to remove the `manifest.xml` as `python_orocos_kdl` is not used by Rock if I am not mistaken (@doudou?).

doudou: > In this particular case it should be fine to remove the manifest.xml as python_orocos_kdl is not used by Rock if I am not mistaken (@doudou?).

I don't for sure (only use KDL C++). Not sure about the rest of the Rock users. I'll drop a line on the mailing list.

meyerj:
I assume the new dependency `<test_depend>python-psutil</test_depend>` added in #216 would also require some special treatment for Python 3 now?

sloretz:
Closing, no longer necessary for ROS Noetic (https://github.com/orocos/orocos_kinematics_dynamics/issues/218)

PR #226, opened 2020-03-10
python_orocos_kdl requires orocos_kdl

sloretz:
Noticed by @cottsay. Adds `REQUIRED` so CMake fails at configure time rather than build time.


meyerj:
Thanks, LGTM!

PR #228, opened 2020-03-13
Prevent segfault in PyKDL when asking for nonexisting segment of chain

DriesDM:
Added methodcode to the Chain class to throw an IndexError when trying to retrieve a segment that doesn't exist.


python_orocos_kdl/PyKDL/kinfam.sip
```diff
@@ -251,8 +251,15 @@ public:
...
-        
-    const Segment& getSegment(unsigned int nr)const /Factory/;
+
+    const Segment* getSegment(unsigned int nr)const /Factory/;
+%MethodCode
+    if (a0 < 0 || a0 >= (unsigned int)sipCpp->getNrOfSegments()) {
+        PyErr_SetString(PyExc_IndexError, "Chain index out of range");
+        return 0;
+    }
+    sipRes = &(sipCpp->getSegment(a0));
```
meyerj: Actually I am not sure if this is valid in SIP. The `/Factory/` annotation indicates that the returned pointer is owned by Python, but actually the returned pointer must not be deleted independently. I wonder whether that annotation was actually correct even for the original version of the method.

Why did you change the return type from `const Segment&` to `const Segment*`?

Could anybody who is more proficient with SIP could jump in (@smits)?

DriesDM: I did not mean to change the return type, that must have slipped in while I was testing, good catch.

python_orocos_kdl/PyKDL/kinfam.sip
```diff
@@ -251,8 +251,15 @@ public:
...
     unsigned int getNrOfJoints()const;
     unsigned int getNrOfSegments()const;
-        
-    const Segment& getSegment(unsigned int nr)const /Factory/;
+
+    const Segment* getSegment(unsigned int nr)const /Factory/;
+%MethodCode
+    if (a0 < 0 || a0 >= (unsigned int)sipCpp->getNrOfSegments()) {
+        PyErr_SetString(PyExc_IndexError, "Chain index out of range");
+        return 0;
```
MatthijsBurgh, line 259: Return is also not allowed in sip method code

python_orocos_kdl/PyKDL/pybind11/kinfam.cpp
```diff
@@ -175,8 +175,12 @@ void init_kinfam(pybind11::module &m)
     chain.def("addChain", &Chain::addChain);
     chain.def("getNrOfJoints", &Chain::getNrOfJoints);
     chain.def("getNrOfSegments", &Chain::getNrOfSegments);
-    chain.def("getSegment", (Segment& (Chain::*)(unsigned int)) &Chain::getSegment);
-    chain.def("getSegment", (const Segment& (Chain::*)(unsigned int) const) &Chain::getSegment);
+    chain.def("getSegment", [](const Chain &c, unsigned int nr)
```
meyerj, line 178: I am not so familiar with PyBind11, but this lambda function returns a `const Segment &` unconditionally, because the input is a `const Chain &`. Does it make a difference? Is the returned `Segment` supposed to have reference semantics or does const-correctness matter at all for the Python bindings?

The previous version had a const and a non-const overload, and the SIP version below only has a non-const implementation of `getSegment(nr)` and the const declaration has been removed.

From [this document in the `pybind11` repo](https://github.com/pybind/pybind11/blob/064a03a49be23d5b19109ca4910bf6ebdd272354/docs/limitations.rst) I understand that `const` on returned values is even cast away by `pybind11` and the return `Segment` instance is still mutable in Python, with reference semantics? In that case, would
```suggestion
    chain.def("getSegment", [](Chain &c, unsigned int nr)
```
have worked here, too, to emphasize that the returned `Segment` is mutable and the `Chain` instance can be modified through this method?

meyerj:
It should be noted that this change is not consistent with the C++ version of `getSegment()`, which deliberately does not implement any checks on the index, probably for performance reasons. In C++ it is up to the caller to make sure that the given index is valid.

https://github.com/orocos/orocos_kinematics_dynamics/blob/a2a2dff5eae7671c2d8db23d1173f6bb3cff6a29/orocos_kdl/src/chain.hpp#L78-L96

MatthijsBurgh:
@meyerj Do you think we should implement such checks for all index functions in PyKDL? I will work on a new PR as this one is outdated.

PR #247, opened 2020-05-19
Expose Jacobian Dot Solver via pybind

JohnSilverFarmer:
Exposes `ChainJntToJacDotSolver` to python via pybind.

### DIFF
--- a/orocos_kdl/tests/jacobiandottest.cpp
+++ b/orocos_kdl/tests/jacobiandottest.cpp
@@ -200,11 +200,11 @@ double compare_Jdot_Diff_vs_Solver(const Chain& chain,const double& dt,const int
     MultiplyJacobian(jdot_by_diff,qdot,jdot_qdot_by_diff);
     
     if(verbose){
-        std::cout << "Jdot diff:" << std::endl << jdot_by_diff << std::endl
-                  << "Jdot solver:" << std::endl << jdot_by_solver << std::endl
-                  << "Error:" << std::endl
-                  << jdot_qdot_by_diff-jdot_qdot_by_solver << std::endl
-                  << q << std::endl
+        std::cout << "Jdot diff:\n" << jdot_by_diff << "\n"
+                  << "Jdot solver:\n" << jdot_by_solver << "\n"
+                  << "Error:\n"
+                  << jdot_qdot_by_diff-jdot_qdot_by_solver << "\n"
+                  << q << "\n"
                   << qdot <<std::endl;
     }
     double err = jdot_qdot_by_diff.vel.Norm() - jdot_qdot_by_solver.vel.Norm()
@@ -235,10 +235,10 @@ double compare_d2_Jdot_Symbolic_vs_Solver(bool verbose)
     MultiplyJacobian(jdot_sym,qdot,jdot_qdot_sym);
     
     if(verbose){
-        std::cout << "Jdot symbolic:" << std::endl << jdot_sym <<std::endl
-                  << "Jdot solver:" << std::endl << jdot_by_solver <<std::endl
-                  << "Error:" << std::endl << jdot_qdot_sym-jdot_qdot_by_solver << std::endl
-                  << q << std::endl
+        std::cout << "Jdot symbolic:\n" << jdot_sym << "\n"
+                  << "Jdot solver:\n" << jdot_by_solver << "\n"
+                  << "Error:\n" << jdot_qdot_sym-jdot_qdot_by_solver << "\n"
+                  << q << "\n"
                   << qdot << std::endl;
     }
     double err = jdot_qdot_sym.vel.Norm() - jdot_qdot_by_solver.vel.Norm()
@@ -264,8 +264,8 @@ bool runTest(const Chain& chain,const int& representation)
             success &= err<=eps_diff_vs_solver;
 
             if(!success || print_err){
-                std::cout << "dt: "<< dt << std::endl
-                          << "err: "<< err << std::endl
+                std::cout << "dt: "<< dt << "\n"
+                          << "err: "<< err << "\n"
                           << "eps_diff_vs_solver: " << eps_diff_vs_solver << std::endl;
                 if(!success)
                     break;
@@ -321,7 +321,7 @@ void JacobianDotTest::testD2Symbolic(){
         success &= err_d2_sym <= eps_sym_vs_solver;
         
         if(!success || print_err){
-            std::cout << "err_d2_sym: " << err_d2_sym << std::endl
+            std::cout << "err_d2_sym: " << err_d2_sym << "\n"
             << "eps_sym_vs_solver: "<< eps_sym_vs_solver <<std::endl;
             if(!success)
                 break;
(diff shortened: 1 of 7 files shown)


MatthijsBurgh:
@JohnSilverFarmer I extended the interface fixed any difference between `sip` and `pybind`. Please check my changes.
Could you please add a test to `kinfamtest.py`? Please reuse chain, but also have a look at `jacobiandottest.cpp`, to include the multiple representations in the test.

JohnSilverFarmer:
I added a test similar to the one in `jacobiandottest.cpp`. I was a bit unsure though how to fit it in nicely (formatting etc.) and using the eps from the cpp test it fails quite often in my testing.

MatthijsBurgh:
@JohnSilverFarmer As we are still supporting Python2, please don't use f-strings

MatthijsBurgh:
@JohnSilverFarmer The issue was you used a wider range of random poses, velocity than the cpp test. I converted it to `random.random`, which produces the same output as the random in `kdl`. Now the python test runs fine with the same error margins as cpp.

MatthijsBurgh:
@JohnSilverFarmer Please review all my changes.

JohnSilverFarmer:
> @JohnSilverFarmer The issue was you used a wider range of random poses, velocity than the cpp test. I converted it to `random.random`, which produces the same output as the random in `kdl`. Now the python test runs fine with the same error margins as cpp.

Ah I see - thanks for fixing! Everything looks good to me.

MatthijsBurgh:
Waiting for an approval from @smits or @meyerj, before merging.

orocos_kdl/tests/jacobiandottest.cpp
```diff
@@ -200,10 +200,12 @@ double compare_Jdot_Diff_vs_Solver(const Chain& chain,const double& dt,const int
...
-        std::cout << "Jdot diff : \n" << jdot_by_diff<<std::endl;
-        std::cout << "Jdot solver:\n"<<jdot_by_solver<<std::endl;
-        
-        std::cout << "Error : " <<jdot_qdot_by_diff-jdot_qdot_by_solver<<q<<qdot<<std::endl;
+        std::cout << "Jdot diff:" << std::endl << jdot_by_diff << std::endl
+                  << "Jdot solver:" << std::endl << jdot_by_solver << std::endl
+                  << "Error:" << std::endl
+                  << jdot_qdot_by_diff-jdot_qdot_by_solver << std::endl
+                  << q << std::endl
+                  << qdot <<std::endl;
```
meyerj, line 208: nit: `std::endl` is not equivalent to `"\n"`, because [`std::endl`](https://en.cppreference.com/w/cpp/io/manip/endl) implies a subsequent [flush](http://www.cplusplus.com/reference/ostream/ostream/flush/) operation and might be less performant than `"\n"` for intermediate line breaks. Both or platform-independent as far as I know.

So if you don't want that other processes' output could be printed between the label and the actual value, it would be preferable to keep `"\n"` to insert a line break, or to even only add a single `std::endl` at the very end. Given that the buffers are large enough, the whole block is then written out at once, instead line by line.

In this particular case the difference is mainly aesthetic.

For reference:
- https://gcc.gnu.org/onlinedocs/libstdc++/manual/streambufs.html#io.streambuf.buffering
- https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rio-endl

MatthijsBurgh, line 208:
```suggestion
        std::cout << "Jdot diff:\n" << jdot_by_diff << "\n"
                  << "Jdot solver:\n" << jdot_by_solver << "\n"
                  << "Error:\n"
                  << jdot_qdot_by_diff-jdot_qdot_by_solver << "\n"
                  << q << "\n"
                  << qdot <<std::endl;
```

orocos_kdl/tests/jacobiandottest.cpp
```diff
@@ -253,16 +257,16 @@ bool runTest(const Chain& chain,const int& representation)
...
 
             success &= err<=eps_diff_vs_solver;
 
             if(!success || print_err){
-                std::cout<<" dt:"<< dt<<" err:"<<err
-                <<" eps_diff_vs_solver:"<<eps_diff_vs_solver
-                <<std::endl;
+                std::cout << "dt: "<< dt << std::endl
+                          << "err: "<< err << std::endl
+                          << "eps_diff_vs_solver: " << eps_diff_vs_solver << std::endl;
```
MatthijsBurgh, line 269:
```suggestion
                std::cout << "dt: "<< dt << "\n"
                          << "err: "<< err << "\n"
                          << "eps_diff_vs_solver: " << eps_diff_vs_solver << std::endl;
```

orocos_kdl/tests/jacobiandottest.cpp
```diff
@@ -233,9 +235,11 @@ double compare_d2_Jdot_Symbolic_vs_Solver(bool verbose)
...
     
     if(verbose){
-        std::cout << "Jdot symbolic : \n" << jdot_sym<<std::endl;
-        std::cout << "Jdot solver:\n"<<jdot_by_solver<<std::endl;
-        std::cout << "Error : " <<jdot_qdot_sym-jdot_qdot_by_solver<<q<<qdot<<std::endl;
+        std::cout << "Jdot symbolic:" << std::endl << jdot_sym <<std::endl
+                  << "Jdot solver:" << std::endl << jdot_by_solver <<std::endl
+                  << "Error:" << std::endl << jdot_qdot_sym-jdot_qdot_by_solver << std::endl
+                  << q << std::endl
+                  << qdot << std::endl;
```
MatthijsBurgh, line 242:
```suggestion
        std::cout << "Jdot symbolic:\n" << jdot_sym << "\n"
                  << "Jdot solver:\n" << jdot_by_solver << "\n"
                  << "Error:\n" << jdot_qdot_sym-jdot_qdot_by_solver << "\n"
                  << q << "\n"
                  << qdot << std::endl;
```

orocos_kdl/tests/jacobiandottest.cpp
```diff
@@ -303,24 +307,24 @@ void JacobianDotTest::testKukaDiffBodyFixed(){
...
+        err_d2_sym = compare_d2_Jdot_Symbolic_vs_Solver(verbose);
         
-        success &= err_d2_sym<=eps_sym_vs_solver;
+        success &= err_d2_sym <= eps_sym_vs_solver;
         
         if(!success || print_err){
-            std::cout <<" err_d2_sym:"<<err_d2_sym
-            <<" eps_sym_vs_solver:"<<eps_sym_vs_solver<<std::endl;
+            std::cout << "err_d2_sym: " << err_d2_sym << std::endl
+            << "eps_sym_vs_solver: "<< eps_sym_vs_solver <<std::endl;
```
MatthijsBurgh, line 325:
```suggestion
            std::cout << "err_d2_sym: " << err_d2_sym << "\n"
            << "eps_sym_vs_solver: "<< eps_sym_vs_solver <<std::endl;
```

MatthijsBurgh:
@JohnSilverFarmer Thanks for the contribution

PR #255, opened 2020-06-06
Make .pc file relocatable and refactor cmake config

### LINKED ISSUE #251: make .pc file relocatable
By avoiding to embed the absolute path of the install prefix in e.g. the .pc file. Similar to ros2/urdfdom#20.

Related to ros2/ros2#606.

meyerj:
Fixes #251:
> By avoiding to embed the absolute path of the install prefix in e.g. the .pc file. Similar to ros2/urdfdom#20.

and also addresses https://github.com/orocos/orocos_kinematics_dynamics/issues/251#issuecomment-639401124:
> Using `../../..` in the exposed `INCLUDE_DIRS` adds the standard `include` path not specific to KDL at least twice to the compiler command-line, because it is not recognized as a duplicate, and causes problems with include directory ordering or with treating include directories of upstream packages as system include dirs.
>
> Furthermore forwarding include directories of dependencies like this is not nice and can also break relocation. Instead, `KDLConfig.cmake` should find `Boost` and `Eigen` again with `find_package()`, and not hard-code the include directories found at compile-time.

Tested in Ubuntu 18.04 with CMake 3.10.2 and also in Ubuntu 12.04 with CMake 2.8.7 (with Eigen `3.0.5-1`). In both cases downstream packages find `orocos_kdl` in relocated installation directories with both, CMake and pkg-config.


MatthijsBurgh:
@meyerj PyKDL fails to build because of a issue finding boost. Which tracebacks to `/usr/local/share/orocos_kdl/cmake/orocos_kdl-config.cmake`

meyerj:
> @meyerj PyKDL fails to build because of a issue finding boost. Which tracebacks to `/usr/local/share/orocos_kdl/cmake/orocos_kdl-config.cmake`

Should be fixed by b2cf207. Boost is only required in downstream packages if `KDL_USE_NEW_TREE_INTERFACE` was enabled at build-time. Before, `@Boost_INCLUDE_DIRS@` was just empty in this case.

orocos_kdl/CMakeLists.txt
```diff
@@ -110,15 +108,23 @@ export(TARGETS orocos-kdl
...
 INSTALL(FILES ${PROJECT_BINARY_DIR}/orocos_kdl-config.cmake DESTINATION share/orocos_kdl/cmake)
 INSTALL(FILES ${PROJECT_BINARY_DIR}/orocos_kdl-config-version.cmake DESTINATION share/orocos_kdl/cmake)
 INSTALL(EXPORT OrocosKDLTargets DESTINATION share/orocos_kdl/cmake)
 
+# Generate pkg-config package configuration
+CONFIGURE_FILE(orocos_kdl.pc.in ${CMAKE_CURRENT_BINARY_DIR}/orocos-kdl.pc @ONLY)
+CONFIGURE_FILE(orocos_kdl.pc.in ${CMAKE_CURRENT_BINARY_DIR}/orocos_kdl.pc @ONLY)
+
+INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/orocos-kdl.pc DESTINATION lib${LIB_SUFFIX}/pkgconfig)
+INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/orocos_kdl.pc DESTINATION lib${LIB_SUFFIX}/pkgconfig)
```
MatthijsBurgh, line 127: Should this happen twice for both lines of code?

meyerj: It used to be like this before (moved from `src/CMakeLists.txt`), that two .pc files are created and installed, one as `orocos-kdl.pc` and one as `orocos_kdl.pc`. I assume that the original package name was `orocos-kdl`, but since ROS does not support package names with dashes (`-`) the names with underscore (`_`) have been added later, in order to be compatible to rosbuild and/or catkin as build tools. For non-ROS packages using dashes as separators is actually more common.

The two `install()` commands could be merged into one, or `configure_file()` could be called only once and the same file installed twice with `install(... RENAME)`. Is that what you meant?

MatthijsBurgh: I didn't noticed the dash and the underscore difference.

I would leave it the way it is.

PR #278, opened 2020-08-03
Non-standard M_PI -> orocos' own PI for Mingw

poelmanc:
While M_PI is absurdly common, it's not officially defined in the C++ standard. That may be why orocos uses PI in most places?

These uses of M_PI cause build errors on Mingw with clang, so this changes them to PI.


MatthijsBurgh:
I am fine by not using M_PI. As right now we mix PI and M_PI. If we change anything apply the change in the entire library. So please replace M_PI everywhere. And make sure you include `utilities/utility.h` everywhere it is used.

ahoarau:
Can't believe this is still a problem in 2020 :smile:

MatthijsBurgh:
Closing in favor of #282

PR #280, opened 2020-08-07
Vereshchagin Solver: Adding documentation and improvements on interfaces, computations and test cases

DjoleMNE:
* Extensive documentation for the Vereshchagin solver added:
 
  - The core idea behind solver's derivation and motivation explained
  - Input and output interfaces explained with examples
  - Practical considerations and insights about the solver's implementation and usage summarized

* Solver's main CartToJoint interface improved with clearly separated input and out variables
* Additional getter added for the solver's final output:
  - Resulting Cartesian acceleration for each of the robot's segments
  - The magnitude of the constraint forces acting on the end-effector: Lagrange Multiplier
  - Total torque acting on each joint
  - Above information useful for the state- simulation and estimation purposes

* Improved solver's computations: joint effective inertia added in computations of robot dynamics

* Test cases for the solver extended:
  - Consistency test between Vereshchagin solver and FD (inverse-inertia-based) solver added
  - Internal consistency test for full Hybrid Dynamics solver's computations added
  - Above mentioned tests also represent extensive examples on how the solver can be used for control and simulation purposes


MatthijsBurgh:
Thanks for the extensive work.

I do't have the time right now to look into this thoroughly.

But a few remarks:
- Make sure all commits have the correct user assigned.
- I don't think the change in class name is necessary, which also breaks the API. So please revert.
- Fix the CI
- Reduce the extensive whitespace in the solvertest

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
```diff
@@ -0,0 +1,206 @@
+
```
MatthijsBurgh, line 1: This is very nice to have, but it is the best if this also is shown in documentation generated by Doxygen. So please check how that can be included nicely. 

Right now I have to idea it is better to move this to the header.

Either way, please check the generated documentation.

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
```diff
@@ -0,0 +1,206 @@
...
+# Documentation on the Vereshchagin Hybrid Dynamics solver
+#### Author of this document: Djordje Vukcevic (GitHub account: `djoleMNE`)
+
+## This solver is also known under the following names:
+* Acceleration Constrained Hybrid Dynamics (ACHD) solver
+* Popov-Vereshchagin solver
+---------------------
+
```
MatthijsBurgh, line 9: White space can be limited. Maybe good idea to run a markdown linter.

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
```diff
@@ -0,0 +1,206 @@
...
+Furthermore, a complete spatial vector of imposed constraint forces can be computed [3], from the following relation: **alpha * nu**.
+
+The reader should note that this **constraint_torque** is the **necessary** control command that a user is supposed to send to robot's joints, to achieve the motion that is computed (resolved) by the Popov-Vereshchagin solver. More specifically, here **constraint_torque** represent solution to the **Inverse Dynamics (ID)** problem. Nevertheless, the reason why a user is not supposed to use the **total_torque** values as the control commands for robot's joints, is the fact that the torque contributions that represent the difference between **total_torque** and **constraint_torque** already exist (act) on robot joints. More specifically, these **additional (residual)** contributions are produced on the joints by the already existing natural forces that act on the system.
+
+On the other side, joint accelerations, namely **q_dotdot** provide solution to the **Forward Dynamics (FD)** problem and these quantities can be used for both control (integrate to joint positions/velocities) and simulation purposes. 
+
```
MatthijsBurgh, line 157: Excessive whitespace

orocos_kdl/src/chainidsolver_vereshchagin.cpp
```diff
@@ -25,12 +25,16 @@
...
+* For more details on this solver, see the documentation in 
+* "chainhdsolver_vereshchagin_doc.md"
+*/
 
 namespace KDL
 {
 using namespace Eigen;
 
-ChainIdSolver_Vereshchagin::ChainIdSolver_Vereshchagin(const Chain& chain_, Twist root_acc, unsigned int _nc) :
+ChainHdSolver_Vereshchagin::ChainHdSolver_Vereshchagin(const Chain& chain_, const Twist root_acc, const unsigned int _nc):
```
MatthijsBurgh, line 37: Change the class name breaks the API. The change is so small that it is not worth it.

Use const reference.

Keep space to colon

orocos_kdl/src/chainidsolver_vereshchagin.cpp
```diff
@@ -43,39 +47,52 @@ ChainIdSolver_Vereshchagin::ChainIdSolver_Vereshchagin(const Chain& chain_, Twis
...
+    d.resize(nj);
+
+    // Extract joint effective inertia from the model and store it in the allocated variable
+    int j = nj - 1;
+    for (int i = ns - 1; i >= 0; i--)
+    {
+        if (chain.getSegment(i).getJoint().getType() != Joint::Fixed)
+        {
+            d(j) = chain.getSegment(i).getJoint().getInertia();
+            j--;
```
MatthijsBurgh, line 61: This means that the first few elements will be empty in case of fixed joints. Is there a more elegant solution possible?

orocos_kdl/src/chainidsolver_vereshchagin.cpp
```diff
@@ -316,24 +340,33 @@ void ChainIdSolver_Vereshchagin::final_upwards_sweep(JntArray &q_dotdot, JntArra
...
+
         s.acc = s.F.Inverse(a_p + s.Z * q_dotdot(j) + s.C);//returns acceleration in link distal tip coordinates. For use needs to be transformed
+
         if (chain.getSegment(i - 1).getJoint().getType() != Joint::Fixed)
             j++;
     }
 }
 
+// Returns Cartesian acceleration of links in robot base coordinates
+void ChainHdSolver_Vereshchagin::getTransformedLinkAcceleration(Twists& xDotdot)
```
MatthijsBurgh, line 360: `xDotdot` is inconsistent. Please use `under_score` naming

orocos_kdl/src/chainidsolver_vereshchagin.cpp
```diff
@@ -316,24 +340,33 @@ void ChainIdSolver_Vereshchagin::final_upwards_sweep(JntArray &q_dotdot, JntArra
...
+{
+    assert(xDotdot.size() == ns + 1);
+    xDotdot[0] = acc_root;
+    for (int i = 1; (unsigned)i < ns + 1; i++)
```
MatthijsBurgh, line 364: How about using an unsigned int...

orocos_kdl/src/chainidsolver_vereshchagin.hpp
```diff
@@ -48,33 +49,41 @@ class ChainIdSolver_Vereshchagin : KDL::SolverI
...
-     * \param chain The kinematic chain to calculate the inverse dynamics for, an internal copy will be made.
-     * \param root_acc The acceleration vector of the root to use during the calculation.(most likely contains gravity)
-     *
+     * \param chain The kinematic chain to calculate the hybrid dynamics for, an internal copy will be made.
+     * \param root_acc The acceleration vector of the root to use during the calculation. (Usually contains gravity)
+     * Note: This solver takes gravity acceleration with opposite sign comparead to the KDL's FD and RNE solvers
+     * \param nc Number of constraints imposed on the robot's end-effector. Max is 6. 
      */
-    ChainIdSolver_Vereshchagin(const Chain& chain, Twist root_acc, unsigned int nc);
+    ChainHdSolver_Vereshchagin(const Chain& chain, const Twist root_acc, const unsigned int nc);
```
MatthijsBurgh, line 57: Same comments as the source file, use const reference.

orocos_kdl/src/chainidsolver_vereshchagin.hpp
```diff
@@ -48,33 +49,41 @@ class ChainIdSolver_Vereshchagin : KDL::SolverI
...
      * @return error/success code
      */
-    int CartToJnt(const JntArray &q, const JntArray &q_dot, JntArray &q_dotdot, const Jacobian& alfa, const JntArray& beta, const Wrenches& f_ext, JntArray &torques);
+    int CartToJnt(const JntArray &q, const JntArray &q_dot, JntArray &q_dotdot, const Jacobian& alfa, const JntArray& beta, const Wrenches& f_ext, const JntArray &ff_torques, JntArray &constraint_torques);
 
     /// @copydoc KDL::SolverI::updateInternalDataStructures
     virtual void updateInternalDataStructures();
+
+    //Returns cartesian acceleration of links in base coordinates
+    void getTransformedLinkAcceleration(Twists& xDotdot);
```
MatthijsBurgh, line 85: Same comment as source file, inconsistent naming.

orocos_kdl/tests/solvertest.hpp
```diff
@@ -52,18 +53,19 @@ class SolverTest : public CppUnit::TestFixture
     void FkVelAndIkVelTest();
     void FkPosAndIkPosTest();
     void VereshchaginTest();
-    void IkSingularValueTest() ;
+    void IkSingularValueTest();
```
MatthijsBurgh, line 56: Revert this whitespace change.

orocos_kdl/tests/solvertest.cpp
```diff
@@ -122,14 +122,14 @@ void SolverTest::setUp()
...
+    // Effective motor inertia is included as joint inertia
+    static const double scale=1.0;
+    static const double offset=0.0;
     static const double inertiamotorA=5.0;      // effective motor inertia kg-m^2
     static const double inertiamotorB=3.0;      // effective motor inertia kg-m^2
     static const double inertiamotorC=1.0;      // effective motor inertia kg-m^2
-    static const double damping=0;
-    static const double stiffness=0;
+    static const double damping=0.0;
+    static const double stiffness=0.0;
```
MatthijsBurgh, line 132: `x.0` is not needed, please revert.

orocos_kdl/tests/solvertest.cpp
```diff
@@ -170,6 +170,59 @@ void SolverTest::setUp()
                                inert7));
     motomansia10dyn.addSegment(Segment(Joint(Joint::None),
                                        Frame(Rotation::Identity(),Vector(0.0,0.0,0.155))));
+
```
MatthijsBurgh, line 173: One empty line is enough

orocos_kdl/tests/solvertest.cpp
```diff
@@ -170,6 +170,59 @@ void SolverTest::setUp()
...
+
+    /** 
+     * KUKA LWR 4 Chain with Dynamics Parameters (for Forward Dynamics and Vereshchagin solver tests)
+     * Necessary test model for the Vereshchagin solver: KDL's implementation of the Vereshchagin solver 
+     * can only work with the robot chains that have equal number of joints and segments.
+     */
+	//joint 1
```
MatthijsBurgh, line 180: Too much, please remove the joint comments and the empty lines.

orocos_kdl/tests/solvertest.cpp
```diff
@@ -197,7 +250,7 @@ void SolverTest::UpdateChainTest()
     ChainDynParam dynparam(chain2, Vector::Zero());
     ChainIdSolver_RNE idsolver1(chain2, Vector::Zero());
     unsigned int nr_of_constraints = 4;
-    ChainIdSolver_Vereshchagin idsolver2(chain2,Twist::Zero(),4);
+    ChainHdSolver_Vereshchagin hdsolver(chain2,Twist::Zero(),4);
```
MatthijsBurgh, line 253: Same as class naming, revert back.

orocos_kdl/tests/solvertest.cpp
```diff
@@ -755,7 +812,170 @@ void SolverTest::FkPosAndIkPosLocal(Chain& chain,ChainFkSolverPos& fksolverpos,
...
+    // Arm root acceleration (robot's base mounted on an even surface)
+    // Note: Vereshchagin solver takes root acc. with opposite sign comparead to the KDL's FD and RNE solvers
+    Twist root_Acc(Vector(0.0, 0.0, 9.81), Vector(0.0, 0.0, 0.0));
+
+    ChainHdSolver_Vereshchagin vereshchaginSolver(kukaLWR, root_Acc, number_of_constraints);
+    solver_return = vereshchaginSolver.CartToJnt(q, qd, qdd, alpha_unit_force, beta_energy, f_ext, ff_tau, constrant_tau);
+    if (solver_return < 0) std::cout << "KDL: Vereshchagin solver ERROR: " << solver_return << std::endl;
+
+
+
```
MatthijsBurgh, line 951: Too much whitespace

orocos_kdl/tests/solvertest.cpp
```diff
@@ -755,7 +812,170 @@ void SolverTest::FkPosAndIkPosLocal(Chain& chain,ChainFkSolverPos& fksolverpos,
...
+    vereshchaginSolver.getTransformedLinkAcceleration(xDotdot);
+
+    // std::cout << "Segment (including root) acceleration: " << std::endl;
+    // for (int i = 0; i < ns + 1 ; i++)
+    //     std::cout << xDotdot[i] << std::endl;
+
+    // std::cout << "Joint space output: " << std::endl;
+    // std::cout << "qdd:  " << qdd.data.transpose() << std::endl;
+    // std::cout << "ff tau:  " << ff_tau.data.transpose() << std::endl;
+    // std::cout << "constraint tau:  " << constrant_tau.data.transpose() << std::endl;
```
MatthijsBurgh, line 968: Please remove

orocos_kdl/tests/solvertest.cpp
```diff
@@ -771,10 +991,10 @@ void SolverTest::VereshchaginTest()
...
 
     //Acceleration energy at  the end-effector
-    JntArray betha(1); //set to zero
-    betha(0) = 0.0;
-    //betha(1) = 0.0;
-    //betha(2) = 0.0;
+    JntArray beta(1); //set to zero
+    beta(0) = 0.0;
+    //beta(1) = 0.0;
+    //beta(2) = 0.0;
```
MatthijsBurgh, line 997: Please remove

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1425,136 @@ void SolverTest::LDLdecompTest()
...
+    KDL::Vector n(3.0, -4.0, 5.0);
+    KDL::Wrench f_tool(f, n);
+
+    KDL::Wrenches f_ext(ns);
+    for(unsigned int i=0;i<ns;i++)
+        SetToZero(f_ext[i]);
+
+    f_ext[ns - 1] = f_tool;
+
+
```
MatthijsBurgh, line 1495: Too much whitespace

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1425,136 @@ void SolverTest::LDLdecompTest()
...
+    Vector gravity(0.0, 0.0, -9.81);  // base frame (Robot base mounted on an even surface)
+    KDL::ChainFdSolver_RNE FdSolver = KDL::ChainFdSolver_RNE(kukaLWR, gravity);
+
+    // Call FD function
+    ret = FdSolver.CartToJnt(q, qd, ff_tau, f_ext, qdd);
+    if (ret < 0) std::cout << "KDL: forward dynamics ERROR: " << ret << std::endl;
+
+
+
```
MatthijsBurgh, line 1507: Too much whitespace

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1425,136 @@ void SolverTest::LDLdecompTest()
...
+    // Vereshchagin solver expects that external wrenches are expressed w.r.t. robot's base frame
+    fksolverpos.JntToCart(q, end_effector_pose, kukaLWR.getNrOfSegments());
+    f_ext[ns - 1] = end_effector_pose.M * f_tool;
+
+    ChainHdSolver_Vereshchagin constraintSolver(kukaLWR, root_Acc, numberOfConstraints);
+    ret = constraintSolver.CartToJnt(q, qd, q_dd_Ver, alpha, beta, f_ext, ff_tau, constraint_tau);
+    if (ret < 0) std::cout << "KDL: Vereshchagin solver ERROR: " << ret << std::endl;
+
+
+
```
MatthijsBurgh, line 1543: Too much whitespace

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1425,136 @@ void SolverTest::LDLdecompTest()
...
+    CPPUNIT_ASSERT(Equal(q_dd_Ver(2), qdd(2), eps));
+    CPPUNIT_ASSERT(Equal(q_dd_Ver(3), qdd(3), eps));
+    CPPUNIT_ASSERT(Equal(q_dd_Ver(4), qdd(4), eps));
+    CPPUNIT_ASSERT(Equal(q_dd_Ver(5), qdd(5), eps));
+    CPPUNIT_ASSERT(Equal(q_dd_Ver(6), qdd(6), eps));
+
+    // std::cout << "FF torque: " << ff_tau << std::endl;
+    // std::cout << "Matrix-Inverse-FD jnt. acc.: " << qdd.data.transpose() << std::endl;
+    // std::cout << "Vereshchagin jnt. acc.:      " << q_dd_Ver.data.transpose() << std::endl;
+    // std::cout << "Vereshchagin constraint tau:      " << constraint_tau.data.transpose() << std::endl;
```
MatthijsBurgh, line 1557: Please remove

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1425,136 @@ void SolverTest::LDLdecompTest()
...
+
+    return;
+}
```
MatthijsBurgh, line 1560: Fix no newline at EOF

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
DjoleMNE: I see. Thanks for the feedback! 

I do not have so much experience with generating Doxygen docs, but I will for sure look into it.
Regarding the idea of moving doc to the header: the documentation is extensive but it is necessary due to solver's complexity, thus, I think it might be too large for the header. 
 
Another idea would be to move this doc file to "orocos_kdl/doc" and reference it in solver's source and header files.

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
DjoleMNE: I will run markdown linter on it.

orocos_kdl/src/chainidsolver_vereshchagin.cpp
DjoleMNE: Regarding the change of class name: The name was changed because the "id" part in the class name was misleading to users. However, the name of the file still remained "chainidsolver_vereshchagin" due to the necessity of keeping the track of the changes made in the file. I did not want to change the file name before collaborator's review. Nevertheless, I would suggest keeping the name (since it describes the algorithm correctly) and later on, when other points that you have outlined are corrected, I would suggest to rename the file name as well and update components of KDL that uses this file. I 
think there are not many components (other files) that call this class; from the first search, it seems that only the solvertest files import this class.

DjoleMNE: Regrading the const reference: what exactly do you mean by that? The arguments under the constructor are already under const.

orocos_kdl/src/chainidsolver_vereshchagin.cpp
DjoleMNE: The number of elements in the joint rotor inertia ("d" ) variable, i.e. size of it, should only correspond to the number of active (not-fixed) joint. The reasons for that are the following:
* It does not make sense to consider (include in computations) inertia of the joints that do not perform any motion. If we do that, we might corrupt the accuracy of the solver, since the joint rotor inertia is added in the overall inertia variable D.
* In the "downwards_sweep" of the solver, authors (Ruben and Azamat) made the same indexing for the joints. See, for example how the feedforward joint torques are added in 238 line.

DjoleMNE:
> Thanks for the extensive work.
> 
> I do't have the time right now to look into the math.
> 
> But a few remarks:
> 
> * Make sure all commits have the correct user assigned.
> * I don't think the change in class name is necessary, which also breaks the API. So please revert.
> * Fix the CI
> * Reduce the extensive whitespace in the solvertest
> * Make sure your documentation is also available in the generated docs. This might require moving it to the header.

Thanks for the feedback!

Regarding the CI: the problem is with the "updateInternalDataStructures()" method in the solver. Namely, the method only updates the data structures that are dependent on the number of segments (ns), but not those that are dependent on the number of joints (nj), such as the newly introduced joint inertia d(nj). In solvertest.cpp file, after the chain2 model is changed, this method is called and then, due to the above-mentioned reason, the test fails.

orocos_kdl/src/chainidsolver_vereshchagin.cpp
DjoleMNE: Yes. A new, more elegant solution is implemented in a new commit.

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
DjoleMNE: Maybe @smits can give us more ideas for the location of and generating Doxygen documentation since he has a lot of experience with documenting the KDL.

MatthijsBurgh: I have generated the docs locally, but this entire file does not show. I stay with my opinion this should be included in the header.

MatthijsBurgh:
I have rename the files and added a temp solver which inherits from the renamed solver. So the API doesn't break, but we are stay consistent with filenaming.

DjoleMNE:
> I have rename the files and added a temp solver which inherits from the renamed solver. So the API doesn't break, but we are stay consistent with filenaming.

Sounds great! Thanks!

orocos_kdl/src/chainhdsolver_vereshchagin_doc.md
DjoleMNE: No problem. I will do that as well.

DjoleMNE: The documentation is now moved to the header file. I have generated the Doxygen HTML file on my machine to see the results. Try the document on your browser:  https://drive.google.com/file/d/1UvEQMq4j9Se5z7wX7w5siyjMopU068oM/view?usp=sharing to see if it is ok.

Btw, what should I do with the already existing markdown file (chainhdsolver_vereshchagin_doc.md): should I delete it from the pull request or move it under /orocos_kdl/doc directory?

MatthijsBurgh:
I think the separate markdown file can be removed. No need to have this documentation at 2 places. I generated the docs and everything looks correct.

DjoleMNE:
Hi @MatthijsBurgh. Is everything OK now with the pull request?

MatthijsBurgh:
@DjoleMNE I will check some time this week.

orocos_kdl/src/chainhdsolver_vereshchagin.cpp
```diff
@@ -0,0 +1,522 @@
...
+}
+
+// Returns magnitude of the constraint forces acting on the end-effector: Lagrange Multiplier
+void ChainHdSolver_Vereshchagin::getContraintForceMagnitude(Eigen::VectorXd &nu_)
+{
+    assert(nu_.size() == nu.size());
+    nu_ = nu;
+}
+
+/*
```
MatthijsBurgh, line 375: Any reason why these commented functions are still here?

orocos_kdl/src/chainhdsolver_vereshchagin.hpp
```diff
@@ -0,0 +1,533 @@
...
+    //Returns cartesian acceleration of links in base coordinates
+    void getTransformedLinkAcceleration(Twists& x_dotdot);
+
+    // Returns total torque acting on each joint (constraints + nature + external forces)
+    void getTotalTorque(JntArray &total_tau);
+
+    // Returns magnitude of the constraint forces acting on the end-effector: Lagrange Multiplier
+    void getContraintForceMagnitude(Eigen::VectorXd &nu_);
+
+    /*
```
MatthijsBurgh, line 417: See commented in source file.

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1429,122 @@ void SolverTest::LDLdecompTest()
...
+    int ret;
+    double eps=1.e-3;
+    ChainFkSolverPos_recursive fksolverpos(kukaLWR);
+    Frame end_effector_pose;
+
+    /**
+     * Compute the forward dynamics (joint accelearitions given actuator torques) 
+     * using both solvers and test for consistency
+     */
+    std::cout<<"KDL FD (inverse-inertia version) and Vereshchagin Solvers Consistency Test for KUKA LWR 4 robot"<<std::endl;
```
MatthijsBurgh, line 1444:
```suggestion
    std::cout << "KDL FD (inverse-inertia version) and Vereshchagin Solvers Consistency Test for KUKA LWR 4 robot" << std::endl;
```

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1429,122 @@ void SolverTest::LDLdecompTest()
...
+    f_ext[ns - 1] = f_tool;
+
+    // ########################################################################################
+    // Forward Dynamics Solver (inverse-inertia version)
+    Vector gravity(0.0, 0.0, -9.81);  // base frame (Robot base mounted on an even surface)
+    KDL::ChainFdSolver_RNE FdSolver = KDL::ChainFdSolver_RNE(kukaLWR, gravity);
+
+    // Call FD function
+    ret = FdSolver.CartToJnt(q, qd, ff_tau, f_ext, qdd);
+    if (ret < 0) std::cout << "KDL: forward dynamics ERROR: " << ret << std::endl;
```
MatthijsBurgh, line 1504:
```suggestion
    if (ret < 0)
        std::cout << "KDL: forward dynamics ERROR: " << ret << std::endl;
```

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1429,122 @@ void SolverTest::LDLdecompTest()
...
+    ff_tau(5) = 15.0;
+    ff_tau(6) = -10.0;
+
+    // External Wrench acting on the end-effector, expressed in local link coordinates
+    KDL::Vector f(10.0, -20.0, 30.0);
+    KDL::Vector n(3.0, -4.0, 5.0);
+    KDL::Wrench f_tool(f, n);
+
+    KDL::Wrenches f_ext(ns);
+    for(unsigned int i=0;i<ns;i++)
```
MatthijsBurgh, line 1493:
```suggestion
    for(unsigned int i=0 ;i<ns; i++)
```

orocos_kdl/tests/solvertest.cpp
```diff
@@ -1202,3 +1429,122 @@ void SolverTest::LDLdecompTest()
...
+    JntArray q_dd_Ver(nj); // Resultant joint accelerations
+
+    // External Wrench acting on the end-effector, this time expressed in base link coordinates
+    // Vereshchagin solver expects that external wrenches are expressed w.r.t. robot's base frame
+    fksolverpos.JntToCart(q, end_effector_pose, kukaLWR.getNrOfSegments());
+    f_ext[ns - 1] = end_effector_pose.M * f_tool;
+
+    ChainHdSolver_Vereshchagin constraintSolver(kukaLWR, root_Acc, numberOfConstraints);
+    ret = constraintSolver.CartToJnt(q, qd, q_dd_Ver, alpha, beta, f_ext, ff_tau, constraint_tau);
+    if (ret < 0) std::cout << "KDL: Vereshchagin solver ERROR: " << ret << std::endl;
```
MatthijsBurgh, line 1537:
```suggestion
    if (ret < 0)
        std::cout << "KDL: Vereshchagin solver ERROR: " << ret << std::endl;
```

MatthijsBurgh (changes requested):
I don't have the knowledge to check the new algorithms. Maybe @smits @meyerj do have the knowledge for that.

orocos_kdl/src/chainhdsolver_vereshchagin.cpp
DjoleMNE: As I mentioned in the above description of the pull-request and in https://github.com/DjoleMNE/orocos_kinematics_dynamics/blob/29087bde5ee092170e552176a6a2c599e746f37c/orocos_kdl/tests/solvertest.cpp#L951, these getters are added for the solver's final output and this information is useful for the state- simulation and estimation purposes. This algorithm provides useful features for not-only control use-cases but also for the above-mentioned one.

DjoleMNE: If you refer to the functions that are located below the "getContraintForceMagnitude()" method, then for those functions, I think there were used by Azamat Shakhimardanov while performing experiments for his PhD work. Nevertheless, I think the reason why hi and/or @smits left them there is the fact that these functions could be useful for the users in certain robotics use cases since the algorithm provides useful features for control, simulation and state-estimation purposes. I think that the first 9 functions could indeed be frequently used by KDL users, however, for the rest of the functions I cannot say the same, since those are very specific.

DjoleMNE: Nevertheless, I would suggest keeping those functions there. However, I did not test those functions, except for "getLinkCartesianAcceleration()" method which is similar to the "getTransformedLinkAcceleration()" method that I added. 

For them, i) it would be nice to have the appropriate tests in the solvertest.cpp file and ii) do some renaming, since some of the current function names are ambiguous (cf. getLinkCartesianVelocity() vs. getLinkVelocity() method names).

I would suggest a separate pull request for these particular improvements.

DjoleMNE:
> I don't have the knowledge to check the new algorithms. Maybe @smits @meyerj do have the knowledge for that.

@smits, @meyerj, any comment from your side?

MatthijsBurgh (approved):

MatthijsBurgh:
By lack of comments, I am going to merge this one.

PR #282, opened 2020-08-09
Replace M_PI by custom PI

MatthijsBurgh:
Overruling #278


MatthijsBurgh:
@meyerj ping ;)

meyerj (approved):
I was not aware that `M_PI` is non-standard. The patch LGTM! Thanks!

Issue #292, opened 2020-10-22
cannot pass weight matrix to ChainIkSolverPos_LMA() class in PyKDL 


radhen:
I have PyKDL installed in ROS and I am using it with Python 2.7. 

Is it possible to pass my own **weight matrix** to ChainIkSolverPos_LMA solver in PyKDL? Below is the constructor for the ChainIkSolverPos_LMA() class, 

`ChainIkSolverPos_LMA (const KDL::Chain &_chain, const Eigen::Matrix< double, 6, 1 > &_L, double _eps=1E-5, int _maxiter=500, double _eps_joints=1E-15)`

and L is the **weight matrix**. I can pass all the other arguments but not L. 
I am creating a eigen matrix in python like this

`weight_matrix = eigen.Matrix6D.Identity()`

and passing that as L, but I get the following error 

`
  File "allegro_contour_5_pykdl.py", line 79, in __init__
    self.ik_p_LMA = kdl.ChainIkSolverPos_LMA(chain=self.arm_chain, L=self.weight_matrix)
TypeError: arguments did not match any overloaded call:
  ChainIkSolverPos_LMA(chain: Chain, eps: float = 1e-05, _maxiter: int = 500, _eps_joints: float = 1e-15): 'L' is not a valid keyword argument
  ChainIkSolverPos_LMA(ChainIkSolverPos_LMA): 'L' is not a valid keyword argument
`

Is the full constructor for ChainIkSolverPos_LMA( ) class not available in Python 2.7 ?

MatthijsBurgh:
I am not sure if you are using a released version of KDL/PyKDL or the master branch.

But I think this applies to both. Both for the PyBind11 version(only master) and the SIP version(released and master) of PyKDL only the short constructor is available, so matching https://github.com/orocos/orocos_kinematics_dynamics/blob/master/orocos_kdl/src/chainiksolverpos_lma.hpp#L108.

A constructor like https://github.com/orocos/orocos_kinematics_dynamics/blob/master/orocos_kdl/src/chainiksolverpos_lma.hpp#L95 is not (yet) available in PyKDL.

radhen:
Good to know. Thank you :)

PR #294, opened 2020-10-25
Remove the SIP version of PyKDL

MatthijsBurgh:
I don't see any reason to keep supporting the SIP version of PyKDL. As SIP doesn't provide any benefits on top of PyBind, but maintaining both is extra work. PyBind provides some features, which are not included in SIP. Therefore I think PyBind is the way to go.

I.E. I wanted to add the extended constructor of a solver, requested in https://github.com/orocos/orocos_kinematics_dynamics/issues/292. In PyBind this is a very simple solution, but SIP doesn't support EIGEN classes out of the box and a custom constructor needs to be created.


python_orocos_kdl/INSTALL.md
```diff
@@ -0,0 +1,46 @@
...
+   - (Optional) To change the build type, add: `-DCMAKE_BUILD_TYPE=<DESIRED_BUILD_TYPE>`
+   - (Optional) To change the python version (default=2), set `ROS_PYTHON_VERSION` environment variable to either `2` or `3`.
+7. Compile: `make`
+8. Install the python bindings: `sudo make install`
+9. Make sure `LD_LIBRARY_PATH` is set correctly: `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib`. Add this also
+to your `.bashrc`.
+10. Execute `ldconfig`: `sudo ldconfig`
+11. (Optional) Execute tests:
+    - Python 2: `python2 ../tests/PyKDLtest.py`
+    - Python 3: `python2 ../tests/PyKDLtest.py`
```
meyerj, line 42:
```suggestion
    - Python 2: `python2 ../tests/PyKDLtest.py`
    - Python 3: `python3 ../tests/PyKDLtest.py`
```

meyerj (commented):
I agree that maintaining both is likely not necessary and PyBind works well on all systems I tested newer versions with. Also the documentation and installation instructions update are pretty useful. Thanks for your effort!

Are there any known disadvantages of PyBind, i.e.
- Does it require C++11 or higher or newer versions of any dependency than with SIP? If yes, removing SIP basically implies that we drop support for older platforms, at least for `python_orocos_kdl`.
- Has it been tested on macOS or Windows?

I have not reviewed this patch in detail, but it looks straight-forward. The installation instruction are tailored towards Debian/Ubuntu. That could be mentioned, without the need to document how to install dependencies on other platforms in detail.

smits (approved):
Looks good to me.

MatthijsBurgh:
This has not been tested on MacOS and/or Windows. But pybind11 does support MacOS and Windows, https://pybind11.readthedocs.io/en/stable/basics.html?highlight=windows.

Have we ever supported anything outside Linux? If we support other OS, we need to have CI for it.

According to GCC, from 4.8 it support c++11. GCC 4.8 is available in 14.04. So That is the oldest LTS that is running, so I think it is fair to drop support for older OS.

MatthijsBurgh:
@meyerj I will wait for your approval before merging

meyerj (approved):

Issue #296, opened 2020-11-08
make error --python_orocos_kdl


Mr-Xia-6:
```
orocos_kinematics_dynamics/python_orocos_kdl/build$ make -j8
[ 10%] Building CXX object CMakeFiles/python_module_PyKDL.dir/PyKDL/sipPyKDLpart1.cpp.o
[ 20%] Building CXX object CMakeFiles/python_module_PyKDL.dir/PyKDL/sipPyKDLpart2.cpp.o
[ 30%] Building CXX object CMakeFiles/python_module_PyKDL.dir/PyKDL/sipPyKDLpart3.cpp.o
[ 40%] Building CXX object CMakeFiles/python_module_PyKDL.dir/PyKDL/sipPyKDLpart5.cpp.o
/home/yusheng/xzq_ws/src/OMG-Planner/orocos_kinematics_dynamics/python_orocos_kdl/build/PyKDL/sipPyKDLpart2.cpp: In member function ‘void sipChainJntToJacSolver::updateInternalDataStructures()’:
/home/yusheng/xzq_ws/src/OMG-Planner/orocos_kinematics_dynamics/python_orocos_kdl/build/PyKDL/sipPyKDLpart2.cpp:76:10: error: ‘updateInternalDataStructures’ is not a member of ‘KDL::ChainJntToJacSolver’
          ::ChainJntToJacSolver::updateInternalDataStructures();
          ^
/home/yusheng/xzq_ws/src/OMG-Planner/orocos_kinematics_dynamics/python_orocos_kdl/build/PyKDL/sipPyKDLpart2.cpp: In function ‘PyObject* meth_ChainJntToJacSolver_updateInternalDataStructures(PyObject*, PyObject*)’:
/home/yusheng/xzq_ws/src/OMG-Planner/orocos_kinematics_dynamics/python_orocos_kdl/build/PyKDL/sipPyKDLpart2.cpp:135:62: error: ‘class KDL::ChainJntToJacSolver’ has no member named ‘updateInternalDataStructures’
             (sipSelfWasArg ? sipCpp-> ::ChainJntToJacSolver::updateInternalDataStructures() : sipCpp->updateInternalDataStructures());
                                                              ^
[...]

PR #229, opened 2020-03-29
PyKDL: PyBind11 based

MatthijsBurgh:
This PR includes the following:
- PyKDL based on PyBind11, which is an option, `BUILD_PYKDL_PYBIND11`. So SIP is still there, but PyBind11 is the default. PyBind11 is included as submodule.
- Improvements on PyKDL based on SIP
    - Moved `std::vector` support to separate file.
    - Fixed custom methodcode which contained a return. Which isn't allowed in SIP. Mostly `__getitem__` and `__setitem__` functions.
    - Added some extra constructors, `__repr__` and equality operators.
- Improvements on kdl, which were needed to fix issues in PyKDL.
- Extending the PyKDL tests. Mostly on the `Frames` and `FrameVel` classes. (Still a lot to gain)
- Cleanup of kdl cmake. Contained some duplicate and unneeded code.
- Python 3 support. Both for PyBind11 as SIP. I have chose to include the Fixed in c++ as well and replace all uses of None by Fixed. None is still there and points to Fixed, so any usage of None still works. For SIP I have used the changes of #137, but with a slight difference, python 3 as the new default.

Resolves #219
Fixes #227,  fixes #230

PR #160, opened 2018-07-10
Add upper and lower limits for a joint

GuilhermePSC:
The upper and lower limits of a joint declared in an SDF file are useful sometimes, so I added them as attributes (doubles) for the KDL::Joint.

Signed-off-by: Guilherme Carvalho <guilherme.ps.carvalho@gmail.com>

PR #297, opened 2020-11-09
Version check

MatthijsBurgh:
As suggested in https://github.com/orocos/orocos_kinematics_dynamics/issues/296#issuecomment-724181063

I am not sure we should bump to `1.5.0` or `2.0.0`


meyerj (commented):
Thanks, @MatthijsBurgh, for picking this suggestion up so quickly. In https://github.com/orocos/orocos_kinematics_dynamics/issues/296#issuecomment-724181063 I did not mean to actually bump the KDL version to 1.5, only to add a version check. However, there are some other API-breaking open pull requests that might be worth to do that (e.g. https://github.com/orocos/orocos_kinematics_dynamics/pull/160) soon, but not necessarily for `python_orocos_kdl`.

But there have been some updates that break building the Python bindings within 1.4, like the addition of `KDL::Joint::Fixed` in  https://github.com/orocos/orocos_kinematics_dynamics/commit/dce6742a379698bc6d50afedc8c0a4fc4657d493 (https://github.com/orocos/orocos_kinematics_dynamics/pull/229). For C++ users that was fully backwards-compatible, but `python_orocos_kdl` requires a version that defines this enum constant since then.

Extending what I proposed earlier, I therefore suggest to integrate the patch version number when finding `orocos_kdl` in `python_orocos_kdl`:
```cmake
# find a matching version of orocos_kdl
find_package(orocos_kdl ${PROJECT_VERSION} EXACT REQUIRED)
```
and to bump both version numbers to 1.4.1 (and not yet to 1.5.0). Then at least `python_orocos_kdl` fails to find `orocos_kdl` if it is installed from the Debian/Ubuntu package [orocos-kdl](https://packages.debian.org/source/sid/orocos-kdl) where the upstream version 1.4.0 [has been imported two years ago](https://salsa.debian.org/science-team/orocos-kdl/-/commit/29c4dfe1eda4c5cc033d5040cd3953043ca0b7ec).

I am not even sure whether `EXACT` works with only the major.minor part. Maybe it would only match a version with patch number 0?

In the future we would then need to

- bump the minor version number for [ABI-breaking changes](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B) (e.g. modify the signature of an existing C++ method)
- bump the patch version number for all changes that do not break the ABI, but require a synchronized update in `orocos_kdl` and `python_orocos_kdl` (e.g. add a new non-virtual member function to both).

@smits ?

smits (commented):
These changes should not trigger a minor version bump afaiu

smits:
I agree with what @meyerj says.

orocos_kdl/package.xml
```diff
@@ -4,7 +4,7 @@
   schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>orocos_kdl</name>
-  <version>1.4.0</version>
+  <version>1.5.0</version>
```
meyerj, line 7:
```suggestion
  <version>1.4.1</version>
```

python_orocos_kdl/package.xml
```diff
@@ -4,7 +4,7 @@
   schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>python_orocos_kdl</name>
-  <version>1.4.0</version>
+  <version>1.5.0</version>
```
meyerj, line 7:
```suggestion
  <version>1.4.1</version>
```

python_orocos_kdl/doc/conf.py
```diff
@@ -38,16 +40,16 @@
 
 # General information about the project.
 project = u'Orocos KDL python bindings'
-copyright = u'2012, Ruben Smits'
+copyright = u'2020, Ruben Smits'
```
meyerj, line 43: I am not a lawyer, but I think the year in copyright statements is the year of the first publication (or a range of years, like 2012-2020). I would not update it to 2020 or later here and elsewhere.

Seems like many copyright statements have been updated in https://github.com/orocos/orocos_kinematics_dynamics/commit/61f457a8a759c8e1f8a491dbbc9ca97eebb886bf?

Reference:
- https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

MatthijsBurgh: I checked the entire repo for changes since the 1.4 release. I updated this in #298

meyerj (approved):
LGTM, thanks!

PR #299, opened 2020-11-14
Add Industrial CI for ABI check

MatthijsBurgh:
I have added Industrial CI for its ABI check.

It is only executed if the target branch is `release-1.3/1.4`, depending on the ROS distro. This way we have a valid check to determine if we can update a release.

To see the industrial CI does work and the ABI check fails, see https://travis-ci.com/github/orocos/orocos_kinematics_dynamics/builds/201521209

I have created a `release-1.4` branch at the `v1.4.0` tag. @meyerj @smits could one of you enable branch protection on it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ jobs:
       if: branch = release-1.3
     - <<: *industrial-ci
       name: Industrial CI Melodic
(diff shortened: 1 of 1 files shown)


meyerj:
> I have created a `release-1.4` branch at the `v1.4.0` tag. @meyerj @smits could one of you enable branch protection on it?

Branch protection is already enabled for all `release-*` branches.

PR #319, opened 2021-01-30
(actions) fix industrial CI missing variable

MatthijsBurgh:
Industrial CI doesn't allow missing variables anymore since https://github.com/ros-industrial/industrial_ci/commit/ee6afa6cfb80e3d611c2044f3aeac175b61327f9

It also showed that PyKDL requires a dependency on python(3).

The ABI check should fail, because master isn't compatible with the release branches. But the checks fail again in a correct way:
Push: https://github.com/orocos/orocos_kinematics_dynamics/actions/runs/524170381
PR: https://github.com/orocos/orocos_kinematics_dynamics/actions/runs/524170414

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,6 +70,7 @@ jobs:
     name: Industrial CI - ${{ matrix.env.ROS_DISTRO }}
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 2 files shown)


meyerj (approved):
LGTM, thanks!

PR #335, opened 2021-06-06
Fix runtime destination for windows install

Tobias-Fischer:


meyerj (approved):
LGTM! Thanks for the patch!

PR #336, opened 2021-06-06
Support IN_LISTS which is required for pybind11 2.6 and newer

Tobias-Fischer:


MatthijsBurgh:
@Tobias-Fischer Should this be set in pybind11?

If not please extend the comment that this is because of pybind11.

Tobias-Fischer:
@MatthijsBurgh - unfortunately it is not set within pybind11. We are packaging with a system pybind11 in conda-forge (https://github.com/conda-forge/orocos-kdl-feedstock) and I came across this issue there. I modified the comment to include the ref to pybind11.

MatthijsBurgh:
@Tobias-Fischer I will accept this for now. But it would be better if you would create a PR in pybind11, so it is set there.

PR #350, opened 2021-08-19
[PyKDL] remove unused FindEigen3.cmake

MatthijsBurgh:


MatthijsBurgh:
@meyerj @smits Just to be sure, it is safe to delete this cmake file, which seems to be unused?

The only time it could be used, which is not tested in CI, is packaging. But than it should appear in the `CMakeLists.txt`, which it doesn't. So therefore I think it is safe to delete this file.

meyerj (approved):
Yes, indeed, should be safe to remove. This copy of `orocos_kdl/cmake/FindEigen3.cmake` is not used anywhere anymore, and the one in package `orocos_kdl` has been updated in the meantime.

However, Eigen headers are still required to compile `python_orocos_kdl`. But the headers are already added to `${orocos_kdl_INCLUDE_DIRS}` by [`orocos_kdl-config.cmake`](https://github.com/orocos/orocos_kinematics_dynamics/blob/8666d8efc66e07748176af81e730ff5c03fb8faa/orocos_kdl/orocos_kdl-config.cmake.in#L10-L14), and there is no need to `find_package(Eigen3)` in `python_orocos_kdl` separately.

Issue #353, opened 2021-09-06
Fails to make:  parse error in template argument list


maguilara:
Hi, I come from downstream, from the orocos-kdl-python package: https://github.com/acxz/pkgbuilds/issues/145
Running arch linux: 5.10.62-1-lts #1 SMP Fri, 03 Sep 2021 10:00:32 +0000 x86_64 GNU/Linux
Installing ros on arch (using this package: https://aur.archlinux.org/packages/ros-noetic-desktop-full/) requires to install some other "libraries", one of these is your package (https://aur.archlinux.org/packages/orocos-kdl-python-git/)
File is compiled using the following commands (you can tell when each step is happening): https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=orocos-kdl-python-git

I get the following error:

```

==> Making package: orocos-kdl-python-git r1039.79adc6f-1 (Mon 06 Sep 2021 04:19:51 PM -05)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Updating orocos_kinematics_dynamics git repo...
Fetching origin
==> Validating source files with sha256sums...
    orocos_kinematics_dynamics ... Skipped
==> Removing existing $srcdir/ directory...
==> Extracting sources...
  -> Creating working copy of orocos_kinematics_dynamics git repo...
Cloning into 'orocos_kinematics_dynamics'...
done.
==> Starting prepare()...
Submodule 'python_orocos_kdl/pybind11' (https://github.com/pybind/pybind11.git) registered for path 'python_orocos_kdl/pybind11'
Cloning into '/home/myusername/.cache/yay/orocos-kdl-python-git/src/orocos_kinematics_dynamics/python_orocos_kdl/pybind11'...
Submodule path 'python_orocos_kdl/pybind11': checked out '787d2c88cafa4d07fb38c9519c485a86323cfcf4'
==> Starting pkgver()...
==> Updated version: orocos-kdl-python-git r1103.bb2493f-1
==> Sources are ready.
==> Making package: orocos-kdl-python-git r1103.bb2493f-1 (Mon 06 Sep 2021 04:19:58 PM -05)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> WARNING: Using existing $srcdir/ tree
==> Starting pkgver()...
==> Starting build()...
-- The C compiler identification is GNU 11.1.0
-- The CXX compiler identification is GNU 11.1.0
[...]

PR #355, opened 2021-09-08
Get rid of  using Eigen and std namespace

MatthijsBurgh:
Fixes https://github.com/orocos/orocos_kinematics_dynamics/issues/353


meyerj (approved):
LGTM :+1:

The `using namespace` directives in KDL source files and unit tests, which is the majority, should not be a problem. But some of them were removed from header files, too, where they could have unwanted side effects in derived code like in `python_orocos_kdl` (#353). In that case it probably was `using namespace Eigen` in [chainiksolvervel_ping_givens.hpp](https://github.com/orocos/orocos_kinematics_dynamics/pull/355/files?w=1#diff-bcef3caaba648384c51beeb0b4acd9b0d698d5bcb845a59cd74789323dc87c4e).

PR #358, opened 2021-09-28
Add degraded status chainiksolverpos

craigirobot:
The nr position solver currently produces a non-convergence error if the Cartesian pose does not converge to the desired value within the specified maximum number of iterations.  However, the lack of convergence could be due to the singularity avoidance being active leaving the solver in a "degraded" state so that it cannot achieve the desired pose.  This commit adds a degraded flag which is checked when the maximum number of iterations is reached.  If it is degraded, then the solver error is set to E_DEGRADED and the user can then decide whether it should be allowed to continue.  Otherwise, the error is set to E_MAX_ITERATIONS_EXCEEDED which would normally terminate the process.  In addition, functions were added to set/get the input parameters (maxiter,eps) and retrieve other data from the solver (twist error, number of iterations, velocity solver status).

### DIFF
--- a/orocos_kdl/src/chainiksolverpos_nr.cpp
+++ b/orocos_kdl/src/chainiksolverpos_nr.cpp
@@ -28,7 +28,10 @@ namespace KDL
         chain(_chain),nj (chain.getNrOfJoints()),
         iksolver(_iksolver),fksolver(_fksolver),
         delta_q(_chain.getNrOfJoints()),
-        maxiter(_maxiter),eps(_eps)
+        maxiter(_maxiter),
+        eps(_eps),
+        numiter(0),
+        ikvelstatus(E_NOERROR)
     {
     }
 
@@ -48,23 +51,50 @@ namespace KDL
             return (error = E_SIZE_MISMATCH);
 
         q_out = q_init;
+        bool degraded = false;
 
-        unsigned int i;
-        for(i=0;i<maxiter;i++){
+        for(numiter=1;numiter<=maxiter;numiter++){
+            degraded = false;        // prove otherwise
             if (E_NOERROR > fksolver.JntToCart(q_out,f) )
                 return (error = E_FKSOLVERPOS_FAILED);
             delta_twist = diff(f,p_in);
-            const int rc = iksolver.CartToJnt(q_out,delta_twist,delta_q);
-            if (E_NOERROR > rc)
+            ikvelstatus = iksolver.CartToJnt(q_out,delta_twist,delta_q);
+            if (E_NOERROR > ikvelstatus) {
                 return (error = E_IKSOLVER_FAILED);
-            // we chose to continue if the child solver returned a positive
+            }
+            // check for degraded solution (but can still continue motion)
+            else if (E_NOERROR < ikvelstatus) {
+                degraded  = true;
+            }
+            // we choose to continue if the child solver returned a positive
             // "error", which may simply indicate a degraded solution
             Add(q_out,delta_q,q_out);
             if(Equal(delta_twist,Twist::Zero(),eps))
                 // converged, but possibly with a degraded solution
-                return (rc > E_NOERROR ? E_DEGRADED : E_NOERROR);
+                return (ikvelstatus > E_NOERROR ? E_DEGRADED : E_NOERROR);
         }
-        return (error = E_MAX_ITERATIONS_EXCEEDED);        // failed to converge
+
+        // update cartesian output to be in sync with final joint solution
+        (void)fksolver.JntToCart(q_out,f);
+
+        // not converged but singularity avoidance is active so okay
+        if (degraded) {
+            return (error = E_DEGRADED);
+        } else {
+            return (error = E_MAX_ITERATIONS_EXCEEDED);
+        }
+    }
+
+    void ChainIkSolverPos_NR::setEps(const double _eps)
+    {
+        if (0 < _eps) eps = _eps;
+        // else silently ignore
+    }
+
+    void ChainIkSolverPos_NR::setMaxIter(const unsigned int _maxiter)
+    {
+        if (1 <= _maxiter) maxiter = _maxiter;
+        // else silently ignore
     }
 
     ChainIkSolverPos_NR::~ChainIkSolverPos_NR()
--- a/orocos_kdl/src/chainiksolverpos_nr.hpp
+++ b/orocos_kdl/src/chainiksolverpos_nr.hpp
@@ -73,6 +73,47 @@ namespace KDL {
          */
         virtual int CartToJnt(const JntArray& q_init, const Frame& p_in, JntArray& q_out);
 
+        /**
+         * Set maximum number of iterations
+         */
+        void setMaxIter(const unsigned int maxiter_in);
+
+        /**
+         * Set epsilon
+         * \pre 0 < eps, otherwise eps is ignored
+         */
+        void setEps(const double eps_in);
+
+        /**
+         * Get maximum number of iterations
+         * \pre 1 <= maxiter, otherwise maxiter is ignored
+         */
+        unsigned int getMaxIter()const { return maxiter; }
+
+        /**
+         * Get epsilon
+         */
+        double getEps()const { return eps; }
+
+        /**
+         * Get delta twist from last call to CartToJnt()
+         */
+        void getDeltaTwist(KDL::Twist& _delta_twist)
+        {
+            _delta_twist    = delta_twist;
+        }
+
+        /**
+         * Get status of ik velocity solver
+         */
+        int getVelSolverStatus() const { return ikvelstatus; }
+
+        /**
+         * Get number iterations spent in last call to CartToJnt()
+         * Defaults to 0 at construction
+         */
+        unsigned int getNumIters() const { return numiter; }
+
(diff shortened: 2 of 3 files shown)


MatthijsBurgh:
Please follow code styler  more strictly. Please add more documentations what the new variables are used for.

snrkiwi:
> Please follow code styler more strictly. Please add more documentations what the new variables are used for.

@MatthijsBurgh Where is the code style documented? It seems like over the years that no really style has been enforced here. Are there particular things you're looking for now?

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
...
+
+        // update cartesian output to be in sync with final joint solution
+        (void)fksolver.JntToCart(q_out,f);
+
+        // not converged but singularity avoidance is active so okay
+        if (degraded) {
+            return (error = E_DEGRADED);
+        } else {
+            return (error = E_MAX_ITERATIONS_EXCEEDED);
+        }
```
MatthijsBurgh, line 85:
```suggestion
        if (degraded)
            return (error = E_DEGRADED);
        else
            return (error = E_MAX_ITERATIONS_EXCEEDED);
```

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
             return (error = E_SIZE_MISMATCH);
 
         q_out = q_init;
+        bool degraded = false;
 
-        unsigned int i;
-        for(i=0;i<maxiter;i++){
+        for(numiter=1;numiter<=maxiter;numiter++){
+            degraded = false;        // prove otherwise
```
MatthijsBurgh, line 57:
```suggestion
        for(numiter=1; numiter<=maxiter; ++numiter) {
            degraded = false; // prove otherwise
```

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
...
             if (E_NOERROR > fksolver.JntToCart(q_out,f) )
                 return (error = E_FKSOLVERPOS_FAILED);
             delta_twist = diff(f,p_in);
-            const int rc = iksolver.CartToJnt(q_out,delta_twist,delta_q);
-            if (E_NOERROR > rc)
+            ikvelstatus = iksolver.CartToJnt(q_out,delta_twist,delta_q);
+            if (E_NOERROR > ikvelstatus) {
```
MatthijsBurgh, line 62:
```suggestion
            if (E_NOERROR > ikvelstatus)
```

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
...
                 return (error = E_IKSOLVER_FAILED);
-            // we chose to continue if the child solver returned a positive
+            }
+            // check for degraded solution (but can still continue motion)
+            else if (E_NOERROR < ikvelstatus) {
+                degraded  = true;
+            }
```
MatthijsBurgh, line 68:
```suggestion
            // check for degraded solution (but can still continue motion)
            else if (E_NOERROR < ikvelstatus)
                degraded  = true;
```

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
...
+        if (degraded) {
+            return (error = E_DEGRADED);
+        } else {
+            return (error = E_MAX_ITERATIONS_EXCEEDED);
+        }
+    }
+
+    void ChainIkSolverPos_NR::setEps(const double _eps)
+    {
+        if (0 < _eps) eps = _eps;
```
MatthijsBurgh, line 90:
```suggestion
        if (0 < _eps)
            eps = _eps;
```

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,50 @@ namespace KDL
...
+        // else silently ignore
+    }
+
+    void ChainIkSolverPos_NR::setMaxIter(const unsigned int _maxiter)
+    {
+        if (1 <= _maxiter) maxiter = _maxiter;
```
MatthijsBurgh, line 96:
```suggestion
        if (1 <= _maxiter)
            maxiter = _maxiter;
```

orocos_kdl/src/chainiksolverpos_nr.hpp
```diff
@@ -73,6 +73,47 @@ namespace KDL {
...
+         * Set epsilon
+         * \pre 0 < eps, otherwise eps is ignored
+         */
+        void setEps(const double eps_in);
+
+        /**
+         * Get maximum number of iterations
+         * \pre 1 <= maxiter, otherwise maxiter is ignored
+         */
+        unsigned int getMaxIter()const { return maxiter; }
```
MatthijsBurgh, line 91:
```suggestion
        unsigned int getMaxIter() const { return maxiter; }
```

orocos_kdl/src/chainiksolverpos_nr.hpp
```diff
@@ -73,6 +73,47 @@ namespace KDL {
...
+
+        /**
+         * Get epsilon
+         */
+        double getEps()const { return eps; }
```
MatthijsBurgh, line 96:
```suggestion
        double getEps() const { return eps; }
```

orocos_kdl/src/chainiksolverpos_nr.hpp
```diff
@@ -73,6 +73,47 @@ namespace KDL {
...
+
+        /**
+         * Get delta twist from last call to CartToJnt()
+         */
+        void getDeltaTwist(KDL::Twist& _delta_twist)
+        {
+            _delta_twist    = delta_twist;
```
MatthijsBurgh, line 103:
```suggestion
            _delta_twist = delta_twist;
```

orocos_kdl/src/chainiksolverpos_nr.hpp
```diff
@@ -73,6 +73,47 @@ namespace KDL {
...
+
+        /**
+         * Get epsilon
+         */
+        double getEps()const { return eps; }
+
+        /**
+         * Get delta twist from last call to CartToJnt()
+         */
+        void getDeltaTwist(KDL::Twist& _delta_twist)
```
MatthijsBurgh, line 101: Reason why not to just simple return?

MatthijsBurgh (changes requested):

orocos_kdl/src/chainiksolverpos_nr.hpp
craigirobot: Do you mean something like this?

Twist& getDeltaTwist()
{
    return delta_twist;
}

orocos_kdl/src/chainiksolverpos_nr.cpp
```diff
@@ -48,23 +51,49 @@ namespace KDL
             return (error = E_SIZE_MISMATCH);
 
         q_out = q_init;
+        bool degraded = false;
 
-        unsigned int i;
-        for(i=0;i<maxiter;i++){
+        for(numiter=1; numiter<=maxiter; ++numiter) {
```
MatthijsBurgh, line 56:
```suggestion
        for(numiter=1; numiter<=maxiter; ++numiter)
        {
```

orocos_kdl/src/chainiksolverpos_nr.hpp
MatthijsBurgh: Yes, indeed. But then get a const reference.

orocos_kdl/src/chainiksolverpos_nr.cpp
snrkiwi: Wait, didn't you suggest in all other cases that the bracket go on the same line as the statement?

MatthijsBurgh: No brackets should be on a newline.

orocos_kdl/src/chainiksolverpos_nr.hpp
craigirobot: Ah, so this?

const Twist& getDeltaTwist()const
{
    return delta_twist;
}

orocos_kdl/src/chainiksolverpos_nr.cpp
craigirobot: I am confused as well.  Your suggested change above puts the bracket on a new line.

MatthijsBurgh: I get your confusion. I edited my previous comment.

Brackets should be on a newline.

craigirobot: This is fine with me (and preferred), but it is not applied consistently throughout KDL.

orocos_kdl/src/chainiksolverpos_nr.hpp
craigirobot: I just added this commit.

PR #363, opened 2021-10-13
Fix Python bindings for static member functions Frme::DH() and Frame::DH_Craig1989()

meyerj:
Those are static member functions of `KDL::Frame` and hence need to be declared to pybind11 with `def_static()` instead of `def()`:

https://github.com/orocos/orocos_kinematics_dynamics/blob/a789adafa8d09f9d098810b4f9b845fc1a77f389/orocos_kdl/src/frames.hpp#L647-L701

Before:
```py
>>> import PyKDL
>>> PyKDL.Frame().DH_Craig1989(0.,0.,0.,0.)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: DH_Craig1989(): incompatible function arguments. The following argument types are supported:
    1. (self: float, arg0: float, arg1: float, arg2: float) -> PyKDL.Frame

Invoked with: [[           1,           0,           0;
            0,           1,           0;
            0,           0,           1]
[           0,           0,           0]], 0.0, 0.0, 0.0, 0.0
```

With this patch:
```py
>>> import PyKDL
>>> PyKDL.Frame().DH_Craig1989(0.,0.,0.,0.)
[[           1,          -0,           0;
            0,           1,          -0;
            0,           0,           1]
[           0,          -0,           0]]
>>> PyKDL.Frame.DH_Craig1989(0.,0.,0.,0.)
[[           1,          -0,           0;
            0,           1,          -0;
            0,           0,           1]
[           0,          -0,           0]]
>>> 

```

### DIFF
--- a/python_orocos_kdl/tests/framestest.py
+++ b/python_orocos_kdl/tests/framestest.py
@@ -347,19 +347,19 @@ class FramesTestFunctions(unittest.TestCase):
         self.assertEqual(Frame(f).p, f.p)
 
         # Denavit-Hartenberg
-        f = Frame.DH(0.0, pi/2, 0.36, 0.0)
-        self.assertTrue(Equal(
-            Frame.DH(0.0, pi/2, 0.36, 0.0),
-            Frame(Rotation(1, 0, 0,
-                           0, 0, -1,
-                           0, 1, 0),
-                  Vector(0, 0, 0.36))))
-        self.assertTrue(Equal(
-            Frame.DH_Craig1989(0.0, pi/2, 0.36, 0.0),
-            Frame(Rotation(1, 0, 0,
-                           0, 0, -1,
-                           0, 1, 0),
-                  Vector(0, -0.36, 0))))
+        f_dh = Frame(Rotation(1, 0, 0,
+                              0, 0, -1,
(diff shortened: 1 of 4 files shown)


python_orocos_kdl/tests/framestest.py
```diff
@@ -346,6 +346,21 @@ def testFrame(self):
         self.assertEqual(Frame(f).M, f.M)
         self.assertEqual(Frame(f).p, f.p)
 
+        # Denavit-Hartenberg
+        f = Frame.DH(0.0, pi/2, 0.36, 0.0)
```
MatthijsBurgh, line 350: This variable seems to be unused

MatthijsBurgh (changes requested):
I think we should both test calling the method on the class and on an instance.

As in python, the added test succeeds with and without the patch. While calling the method on an instance fails (which you showed in your comment), which is not tested right now.

MatthijsBurgh:
@meyerj have you been able to check my last commit?

meyerj:
> @meyerj have you been able to check my last commit?

Yes, all fine for me. Thanks!

PR #375, opened 2022-01-07
Skip building pybind11 if already available on system

jacobperron:
For example, ROS 2 currently ships with a version of pybind11 that we can use instead of building another version from source.

### DIFF
--- a/python_orocos_kdl/CMakeLists.txt
+++ b/python_orocos_kdl/CMakeLists.txt
@@ -46,7 +46,13 @@ else()
   set(PYTHON_MODULE_EXTENSION ".so")
 endif()
 
-add_subdirectory(pybind11)
+find_package(pybind11 QUIET)
+if(NOT ${pybind11_FOUND})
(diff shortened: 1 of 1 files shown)


python_orocos_kdl/CMakeLists.txt
```diff
@@ -46,7 +46,13 @@ else()
   set(PYTHON_MODULE_EXTENSION ".so")
 endif()
 
-add_subdirectory(pybind11)
+find_package(pybind11 QUIET)
+if(NOT ${pybind11_FOUND})
+  message(STATUS "pybind11 not found, building from source")
+  add_subdirectory(pybind11)
+else()
+  message(STATUS "pybind11 found, building against found version")
```
MatthijsBurgh, line 54:
```suggestion
  message(STATUS "pybind11 found, building against installed version")
```

MatthijsBurgh (changes requested):
Just a small change in a message.

MatthijsBurgh (approved):

PR #389, opened 2022-03-29
Add files via upload

21Ansh:
Fixed build errors  No such file or directory  #include <Eigen/Core>

### DIFF
--- a/orocos_kdl/src/articulatedbodyinertia.cpp
+++ b/orocos_kdl/src/articulatedbodyinertia.cpp
@@ -21,7 +21,7 @@
 
 #include "articulatedbodyinertia.hpp"
 
-#include <Eigen/Core>
+#include <eigen3/Eigen/Core>
 
 namespace KDL{
     
--- a/orocos_kdl/src/articulatedbodyinertia.hpp
+++ b/orocos_kdl/src/articulatedbodyinertia.hpp
@@ -27,7 +27,7 @@
 #include "rotationalinertia.hpp"
 #include "rigidbodyinertia.hpp"
 
-#include <Eigen/Core>
+#include <eigen3/Eigen/Core>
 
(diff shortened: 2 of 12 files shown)


meyerj:
Eigen [is meant to be used with the original include directives](https://eigen.tuxfamily.org/dox/GettingStarted.html), without the `eigen3/` prefix. Instead, when building KDL, the installation of Eigen is normally found by CMake and added to the include path [here](https://github.com/orocos/orocos_kinematics_dynamics/blob/11d8bd62128092ff64ab20e419b295ca72aada32/orocos_kdl/CMakeLists.txt#L52-L57):

https://github.com/orocos/orocos_kinematics_dynamics/blob/11d8bd62128092ff64ab20e419b295ca72aada32/orocos_kdl/CMakeLists.txt#L52-L57

It is also exported in the [generated CMake config](https://github.com/orocos/orocos_kinematics_dynamics/blob/11d8bd62128092ff64ab20e419b295ca72aada32/orocos_kdl/orocos_kdl-config.cmake.in#L25-L31) for package `orocos_kdl` and in [`orocos_kdl.pc`](https://github.com/orocos/orocos_kinematics_dynamics/blob/11d8bd62128092ff64ab20e419b295ca72aada32/orocos_kdl/orocos_kdl.pc.in#L10) for [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) via `Cflags: -I${includedir} @KDL_CFLAGS@`.

If you use neither of both, then indeed you would have to make sure that `/usr/include/eigen3` (or another installation of Eigen3) is in the compiler's include path for all derived packages and/or projects in another way.

> Fixed build errors No such file or directory #include <Eigen/Core>

That does not sound like compiler error? What is the exact error message you are facing?

Issue #384, opened 2022-02-14
Why is nr_of_jnts a double and not unsigned int?


guru-florida:
The constructor for TreeIkSolverPos_Online takes `nr_of_jnts` argument but it is a double. Looking at the source code, the argument is only used to construct the JntArray objects and that constructor takes an `unsigned int`...it seems to me the `nr_of_jnts` argument should also be an unsigned int. I can't see a use case for having non-integer number of joints. The argument is not used anywhere else but the constructor.

https://github.com/orocos/orocos_kinematics_dynamics/blob/4f705fe9e1533802578b5275ddad02b065b2be4e/orocos_kdl/src/treeiksolverpos_online.cpp#L29

I can do a PR if you like.

MatthijsBurgh:
@guru-florida thanks for your comment.

I agree with you. Internally an `uint` is used. So this API doesn't make any sense.

Changing this API, would be a breaking change. Therefore I don't want to change it for now. So I created a milestone for the next breaking release. This could be long way down the road. But I will add this issue to that milestone.

Issue #239, opened 2020-05-14
Inconsistent solver test results


MatthijsBurgh:
https://travis-ci.org/github/orocos/orocos_kinematics_dynamics/jobs/686495037#L1271

FAIL: testFkPosAndIkPos (kinfamtest.KinfamTestFunctions)
1272----------------------------------------------------------------------
1273Traceback (most recent call last):
1274  File "/home/travis/build/orocos/orocos_kinematics_dynamics/python_orocos_kdl/tests/kinfamtest.py", line 224, in testFkPosAndIkPos
1275    self.assertEqual(q, q_solved)
1276AssertionError: [     1.60555     1.57944   -0.248299    -1.48487     3.10299    -1.20087] != [     1.60554     1.57944   -0.248286    -1.48487     3.10301    -1.20088]

MatthijsBurgh:
https://travis-ci.org/github/orocos/orocos_kinematics_dynamics/jobs/688782918
```
3: Test command: /home/travis/build/orocos/orocos_kinematics_dynamics/orocos_kdl/build/tests/solvertest
3: Test timeout computed to be: 10000000
3: ...square problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: underdetermined problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: overdetermined problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: overdetermined problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: .square problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: underdetermined problem
3: KDL-SVD-HouseHolder
3: KDL-SVD-Givens
3: F.0.000000          0.000228      0.523108       0.045580     -0.098212       4.558028      -9.821186     0.410223      0.568421
3: 0.010000          0.000777      0.521924       0.064329     -0.138607       1.874883      -4.039482     0.267959      0.371448
3: 0.020000          0.001560      0.520237       0.092266     -0.198784       2.793640      -6.017780     0.317335      0.440332
3: 0.030000          0.002606      0.517985       0.116808     -0.251636       2.454215      -5.285122     0.300258      0.417227
3: 0.040000          0.003901      0.515194       0.142247     -0.306397       2.543890      -5.476142     0.306241      0.426347
[...]

PR #390, opened 2022-04-01
Fix build warnings on Windows

jacobperron:
We noticed these warnings when building orocos_kdl as part of ROS 2 CI, see https://ci.ros2.org/job/ci_windows/16708/msbuild

I've triggered a new Windows build with the changes in this PR: [Build Status](http://ci.ros2.org/job/ci_windows/16877/)

### DIFF
--- a/orocos_kdl/src/tree.cpp
+++ b/orocos_kdl/src/tree.cpp
@@ -156,8 +156,8 @@ bool Tree::getChain(const std::string& chain_root, const std::string& chain_tip,
     }
 
     // add the segments from the common frame to the tip frame
-    for (int s=static_cast<int>(parents_chain_tip.size())-1; s>-1; s--){
-        chain.addSegment(GetTreeElementSegment(getSegment(parents_chain_tip[s])->second));
+    for (auto rit=parents_chain_tip.rbegin(); rit != parents_chain_tip.rend(); ++rit){
+        chain.addSegment(GetTreeElementSegment(getSegment(*rit)->second));
     }
     return true;
 }
--- a/orocos_kdl/src/tree.cpp
+++ b/orocos_kdl/src/tree.cpp
@@ -156,7 +156,7 @@ bool Tree::getChain(const std::string& chain_root, const std::string& chain_tip,
     }
 
     // add the segments from the common frame to the tip frame
-    for (int s=parents_chain_tip.size()-1; s>-1; s--){
+    for (int s=static_cast<int>(parents_chain_tip.size())-1; s>-1; s--){
         chain.addSegment(GetTreeElementSegment(getSegment(parents_chain_tip[s])->second));
     }
     return true;
--- a/orocos_kdl/src/treeiksolverpos_online.hpp
+++ b/orocos_kdl/src/treeiksolverpos_online.hpp
@@ -61,7 +61,7 @@ public:
      * @return
      */
 
-    TreeIkSolverPos_Online(const double& nr_of_jnts,
+    TreeIkSolverPos_Online(const unsigned int& nr_of_jnts,
                            const std::vector<std::string>& endpoints,
                            const JntArray& q_min,
                            const JntArray& q_max,
--- a/orocos_kdl/src/treeiksolverpos_online.hpp
+++ b/orocos_kdl/src/treeiksolverpos_online.hpp
@@ -61,7 +61,7 @@ public:
      * @return
      */
 
-    TreeIkSolverPos_Online(const unsigned int& nr_of_jnts,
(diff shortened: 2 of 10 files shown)


orocos_kdl/src/tree.cpp
```diff
@@ -156,7 +156,7 @@ bool Tree::getChain(const std::string& chain_root, const std::string& chain_tip,
     }
 
     // add the segments from the common frame to the tip frame
-    for (int s=parents_chain_tip.size()-1; s>-1; s--){
+    for (int s=static_cast<int>(parents_chain_tip.size())-1; s>-1; s--){
```
MatthijsBurgh, line 159: Why not go for an `unsigned int` here?

orocos_kdl/src/treeiksolverpos_online.hpp
```diff
@@ -61,7 +61,7 @@ class TreeIkSolverPos_Online: public TreeIkSolverPos {
      * @return
      */
 
-    TreeIkSolverPos_Online(const double& nr_of_jnts,
+    TreeIkSolverPos_Online(const unsigned int& nr_of_jnts,
```
MatthijsBurgh, line 64: This is an API breaking change. This has been discussed in https://github.com/orocos/orocos_kinematics_dynamics/issues/384. So I think we should skip this one.

orocos_kdl/src/utilities/error_stack.cxx
```diff
@@ -57,7 +57,11 @@ void IOTracePopStr(char* buffer,int size) {
         *buffer = 0;
         return;
     }
+#if defined(_WIN32)
+    strncpy_s(buffer,size,errorstack.top().c_str(),size);
+#else
     strncpy(buffer,errorstack.top().c_str(),size);
+#endif
```
MatthijsBurgh, line 64: Any reason why not to use `strncpy_s` on all platforms?

MatthijsBurgh (changes requested):

orocos_kdl/src/tree.cpp
jacobperron: The condition for the loop is to check if `s>-1`, so it must be able to become negative.

orocos_kdl/src/utilities/error_stack.cxx
jacobperron: No all libraries (like glibc) define `strncpy_s`. Microsoft's compiler, however, insists on using it :shrug:

orocos_kdl/src/treeiksolverpos_online.hpp
jacobperron: Updated to avoid breaking API: [d70e705](https://github.com/orocos/orocos_kinematics_dynamics/pull/390/commits/d70e705f2f766adc4b22f931a3aaa7ffd470feac)

jacobperron: Not sure why, but it looks like this change ([d70e705](https://github.com/orocos/orocos_kinematics_dynamics/pull/390/commits/d70e705f2f766adc4b22f931a3aaa7ffd470feac)) broke a test :thinking:

jacobperron: Since most of the members we were initializing with this value where being overridden immediately in the constructor, I've just removed them from the initializer list ([fc0dfd8](https://github.com/orocos/orocos_kinematics_dynamics/pull/390/commits/fc0dfd80cbe1810f7e27c10a860c7c1896174f99)). CI looks happy now.

orocos_kdl/src/tree.cpp
MatthijsBurgh: And if we check for `s>=0`¿

jacobperron: If `s` is an `unsigned int` it cannot be negative, so `s--` will cause an integer overflow and this will become an infinite loop.

I've updated the code to use a reverse iterator, which I think makes it more readable ([3324420](https://github.com/orocos/orocos_kinematics_dynamics/pull/390/commits/3324420efff9b58b53d37680beb01a81e440d6f6)).

jacobperron:
I guess CI is flaky? Since the same change passed CI on my fork: https://github.com/jacobperron/orocos_kinematics_dynamics/runs/5796424695?check_suite_focus=true

MatthijsBurgh:
> I guess CI is flaky? Since the same change passed CI on my fork:

Yes, see https://github.com/orocos/orocos_kinematics_dynamics/issues/239

MatthijsBurgh (approved):

PR #391, opened 2022-04-01
Do not pass generated header to add_library

jacobperron:
Newer versions of cmake (>3.22.1) emit the following warning:

  CMake Warning (dev) at src/CMakeLists.txt:69 (ADD_LIBRARY):
    Policy CMP0115 is not set: Source file extensions must be explicit.  Run
    "cmake --help-policy CMP0115" for policy details.  Use the cmake_policy
    command to set the policy and suppress this warning.

I don't know exactly why passing this generated header triggers the warning,
but passing it to add_library doesn't seem necessary.


clalancette:
> I don't know exactly why passing this generated header triggers the warning, but passing it to add_library doesn't seem necessary.

In my experience, that warning comes about when the referenced file doesn't exist.  In this case, I think the problem is that `config.h` in that line actually lives at `${CMAKE_CURRENT_BINARY_DIR}/config.h`.  But I agree with you that it shouldn't be needed at all, so I think this patch is fine.

MatthijsBurgh (approved):

PR #394, opened 2022-05-11
Add hash function to "dataclasses"

MatthijsBurgh:
`std::hash` is only available from c++11. Do we want to to check for this (`#if (__cplusplus > 201103L)`? Or can we assume that all users have minimal c++11?


MatthijsBurgh:
@meyerj @smits friendly ping ;)

PR #409, opened 2022-08-19
do not read outside array

PetervDooren:
An error in the Vereshchagin hybrid dynamics solver causes NaNs to be propagated through the data. The cause is a faulty loop in the downwards sweep which starts by reading segment ns, where ns is the number of segments in the chain. However the chain only contains segments 0 through to ns-1. 

Small fix was implemented. Ideally unit tests could prevent this in the future.


PetervDooren:
Just to clarify for future reference. This was my mistake. I used the model of the Franke emika panda robot which ends with an extra segment. The reason NaNs were propagated through my data was that the final segment had 0 mass. Naturally this leads to an unsolvable dynamics problem.

PR #412, opened 2022-09-30
(PyKDL) let catkin export CMake and PkgConfig files

MatthijsBurgh:
@meyerj You added this in 7e1f40d4d043811ea12ac1f620dae431d1338df1. But this prevents to find the package in a CMakeLists.

Is there any reason why you did add these flags?

### DIFF
--- a/python_orocos_kdl/CMakeLists.txt
+++ b/python_orocos_kdl/CMakeLists.txt
@@ -35,10 +35,7 @@ set(LIBRARY_NAME "PyKDL")
 # catkin-specific configuration (optional)
 find_package(catkin QUIET)
 if(catkin_FOUND)
-  catkin_package(
-    SKIP_CMAKE_CONFIG_GENERATION
-    SKIP_PKG_CONFIG_GENERATION
-  )
+  catkin_package()
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CATKIN_DEVEL_PREFIX}/${PYTHON_INSTALL_DIR})
   set(PYTHON_SITE_PACKAGES_INSTALL_DIR "${PYTHON_INSTALL_DIR}")
 endif()


meyerj:
I do not remember the details, but it was a part of https://github.com/orocos/orocos_kinematics_dynamics/pull/285. In its description I wrote:

> Macro `catkin_package()` takes care of installing `package.xml` and setup scripts. Usually it also installs cmake and pkg-config configuration files, but we skip this because `orocos_kdl` already provides its own configuration and the Python bindings do not require it.
> 
> If the Python module is built in the correct path inside the devel-space, pointed to by `PYTHON_INSTALL_DIR` and `CATKIN_GLOBAL_PYTHON_DESTINATION`, and also installed to this directory, no env-hooks are required to update the `PYTHONPATH` and `python_orocos_kdl` even works in devel-space.

So on the one hand I still think that it is not required, because you would typically not try to find `python_orocos_kdl` in a CMake project, but on the other hand there must have been an issue with that with at least one of the tested build tools. Before https://github.com/orocos/orocos_kinematics_dynamics/pull/285, `python_orocos_kdl` was built as a pure CMake package and finding it in CMake was not possible either.

The package does not provide any linkable C++ library nor installed headers. So what is your use case of writing something like
```cmake
find_package(python_orocos_kdl REQUIRED)
```
or
```cmake
find_package(catkin REQUIRED COMPONENTS python_orocos_kdl)
```
? But if there is one, and if it does not break anything, please feel free to merge this pull request. For Python packages that use `PyKDL` it is sufficient to list `python_orocos_kdl` as a dependency in the `package.xml` manifest without mentioning it in `CMakeLists.txt`.

For package `orocos_kdl` (the C++ library) the two arguments `SKIP_CMAKE_CONFIG_GENERATION` and `SKIP_PKG_CONFIG_GENERATION` are passed to `catkin_package()` because otherwise the generated configuration files would clash with the ones generated by `orocos_kdl/CMakeLists.txt` itself (from [`orocos_kdl-config.cmake.in`](https://github.com/orocos/orocos_kinematics_dynamics/blob/04200414743658d9a0de7132bb475b624bfc2694/orocos_kdl/orocos_kdl-config.cmake.in) and [`orocos_kdl.pc.in`](https://github.com/orocos/orocos_kinematics_dynamics/blob/04200414743658d9a0de7132bb475b624bfc2694/orocos_kdl/orocos_kdl.pc.in)). For that reason, and if there is indeed a valid use case, it probably would be better to also add a custom `python_orocos_kdl-config.cmake.in` template such that it behaves the same, independent of whether it has been built with catkin (for ROS users) or CMake directly (for non-ROS users).

PR #442, opened 2023-06-23
resolved: build error 

raghavvs:
Added missing `SET(CMAKE_CXX_STANDARD 11)` to CMakeLists.txt


MatthijsBurgh:
This PR is wrong in 2 ways.
1. You add some empty `.DS_store` files
2.  The library also works with newer CPP versions. So we shouldn't limit this.

I have no clue about your system, why you need this
But maybe you should just add this as argent to CMake.

PR #493, opened 2025-06-27
(PyKDL) expose ChainExternalWrenchEstimator and ChainHdSolver_Vereshchagin

NRRHHH:
Adds Python bindings for `ChainExternalWrenchEstimator` and `ChainHdSolver_Vereshchagin`, including tests. It also fixes several C++ issues found while binding them.

## PyKDL

- Expose `ChainExternalWrenchEstimator`, including its error codes, so callers can interpret the return value of `JntToExtWrench`.
- Expose `ChainHdSolver_Vereshchagin`. pybind11 copies `std::vector` and Eigen arguments, so writes to a C++ out-parameter never reach the caller's object. Getters with such out-parameters therefore take the pre-sized container and return the filled one.
- Register `ChainDynParam` and `ChainExternalWrenchEstimator` with their `SolverI` base. `getError()` now works and `isinstance(solver, SolverI)` returns `True`.
- Remove numpy as a runtime dependency:
  - `getConstraintForceMagnitude` returns a list of doubles.
  - The weighted `ChainIkSolverPos_LMA` constructor takes `L` as any sequence of 6 doubles. Lists, tuples and numpy arrays are all still accepted, and a wrong length now raises `ValueError` instead of reading past the end of the input.
  - With no Eigen types left in any binding, the `pybind11/eigen.h` include is dropped.
- Port `VereshchaginTest` and `ExternalWrenchEstimatorTest` from `solvertest.cpp`. The KUKA LWR chain with dynamics parameters is added as a shared test fixture. The estimator test uses a fixed seed so any failure can be reproduced.

## orocos_kdl

- `ChainHdSolver_Vereshchagin` now inherits `SolverI` publicly. It was the only solver that didn't, which made `getError()` and `strError()` inaccessible.
- `ChainExternalWrenchEstimator`'s static error constants are now defined out of line, like `ChainJntToJacDotSolver`'s. Before, taking their address failed to link.
- **API change:** `ChainHdSolver_Vereshchagin::getContraintForceMagnitude` is renamed to `getConstraintForceMagnitude`, without an alias for the old spelling.

Also adds a `.gitignore`.


MatthijsBurgh:
Please create a separate commit for the `.gitignore` file

MatthijsBurgh:
@PeterBowman Are you planning to add the tests? I don't want to add features without adding tests.

PeterBowman:
@MatthijsBurgh sure, I was a bit busy, next week I'll get back at this. As I noted before, this class is coupled with several other solvers in [solvertest.cpp](https://github.com/orocos/orocos_kinematics_dynamics/blob/ec5d67e5b04f785d44ac012816376945e7c09e7b/orocos_kdl/tests/solvertest.cpp), which are currently not covered in [kinfamtest.py](https://github.com/orocos/orocos_kinematics_dynamics/blob/ec5d67e5b04f785d44ac012816376945e7c09e7b/python_orocos_kdl/tests/kinfamtest.py). In order to translate ChainExternalWrenchEstimator's C++ test cases to Python, it seems to be necessary to add tests for those other classes, too. Is this correct?

MatthijsBurgh:
When a test case is repeated for multiple solvers, it would be preferable to copy them all to python.

